PR #4, opened 2017-09-08
Adding info commands

WxWatch:
Hi.

I still need to get all the `roleId`s that should be able to add/remove commands (right now it's only Gurus, probably need to add admins/rito/etc), but if someone could look over what I wrote and make sure I'm not missing anything blatantly obvious, 🌮


src/Info.ts
```diff
@@ -0,0 +1,107 @@
...
+// If the Collection contains anything in the other array
+const findOne = (arr1: Discord.Collection<string, Discord.Role>, arr2: Array<any>) => {
+    return arr1.some(v => {
+        return arr2.indexOf(v) >= 0;
+    });
+};
+
+export interface InfoData {
+    Command: string;
+    Message: string;
```
molenzwiebel: My only issue is that these fields are capitalized. TypeScript conventions use lowerCamelCase for field names. I guess that config values elsewhere also use PascalCase, but that's because Q is a goof.

src/Info.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    onBot() {
+        console.log("Info extension loaded.");
+    }
+
+    onInfo(message: Discord.Message) {
+        if (!findOne(message.member.roles, this.sharedSettings.info.allowedRoles)) return;
+
+        const split: string[] = message.cleanContent.split(" ");
```
molenzwiebel: You shouldn't need to have the `: string[]` type assertion here.

src/Info.ts
WxWatch: Yeah, I didn't like capitalizing either, but I was just doing it for consistency. I think we either need to keep it like it is and commit to refactoring all of them later or going back right now and changing all of them.

src/Info.ts
WxWatch: 👍 I'll remove it

src/Info.ts
```diff
@@ -0,0 +1,107 @@
...
+    Message: string;
+}
+
+export default class Info {
+    private bot: Discord.Client;
+    private infos: InfoData[];
+    private sharedSettings: SharedSettings;
+
+    constructor(bot: Discord.Client, sharedSettings: SharedSettings, userFile: string) {
+        console.log("Requested Thinking extension..");
```
Derpthemeus: This should be changed to the extension name

WxWatch: Oh no, you noticed my copy/paste work :(

Querijn:
Thanks!

PR #7, opened 2017-10-02
Don't return null if the `x-app-rate-limit` header doesn't exist

WxWatch:
This should fix the case where Botty is bamboozling us into thinking a key has been reset when there's a 500 error without an `x-app-rate-limit` header

### DIFF
--- a/src/KeyFinder.ts
+++ b/src/KeyFinder.ts
@@ -77,7 +77,16 @@ export default class KeyFinder {
             }
         });
 
-        return resp.status === 403 ? null : resp.headers.get("x-app-rate-limit");
+        // resp.headers.get("x-app-rate-limit") might return null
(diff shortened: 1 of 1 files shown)


Derpthemeus:
Unless I'm missing something, I think this issue was already fixed in 9eaba4a94e2e6504bbd25973cc4934c81ba4c138 by returning `null` if the key yields a 403 response, and `undefined` if it yields a non-403 response without a `x-app-rate-limit` header.

WxWatch:
oh, I thought I saw some keys getting prematurely kicked yesterday, but my days might be all screwed up.

Feel free to close if this is dumb

Querijn:
@Derpthemeus is this ~~~dumb~~~done?

PR #9, opened 2017-10-06
Allows botty to read notes using just the note name (ex: !repo)

s-coimbra21:


Querijn:
Might conflict with other functionality.. It was kind of setup this way on purpose

s-coimbra21:
Yes, ideally, all the builtin commands would be stored as constants in some file and those would be read to disallow already existing commands from being added by `!note add ${cmd}`. But right now only admins can add commands so its not that big of a deal.

Querijn:
You're forgetting we have Derp as admin

s-coimbra21:
`helpers/isDerp.ts`

Querijn:
Thanks!

PR #10, opened 2017-10-15
API Status Implementation

TiFu:
As promised a port of the api status implementation Heimerdinger uses. 

Let me know if you want me to change something.


src/SharedSettings.ts
```diff
@@ -39,5 +39,13 @@ export interface SharedSettings {
...
         server: string,
         channel: string
+    },
+    apiStatus: {
+        checkInterval: number,
+        apiOnFireThreshold: number,
+        statusUrl: string,
+        command: string,
+        aliases: Array<string>,
+        onFireImages: Array<string>
```
molenzwiebel: `string[]` is the normal way to describe an array

src/app.ts
```diff
@@ -13,6 +13,7 @@ import Logger from "./Logger";
 import { fileBackedObject } from "./FileBackedObject";
 import { SharedSettings } from "./SharedSettings";
 import { PersonalSettings } from "./PersonalSettings";
+import ApiStatus from './ApiStatus'
```
molenzwiebel: `"`, semicolon

src/ApiStatusApi.ts
```diff
@@ -0,0 +1,49 @@
+import fetch from "node-fetch";
+
+type RegionState = "up" | "troubled" | "down";
+
+export interface RegionStatus {
+    "uptime": string,
+    "performance": string,
+    "state": RegionState
```
molenzwiebel: Normally we only quote keys if they contain characters that aren't normally allowed (like `-`).

src/ApiStatus.ts
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    private startsWithAlias(msgContent: string): boolean {
+        for (let alias of this.sharedSettings.apiStatus.aliases) {
+            console.log("alias " + alias);
+            if (msgContent.startsWith("!" + alias)) {
+                return true;
+            }
+        }
+        return false;
```
molenzwiebel: This can be simplified to `return this.sharedSettings.apiStatus.aliases.some(x => msgContent.startsWith("!" + x))`. Also, the debug ;)

src/ApiStatus.ts
```diff
@@ -0,0 +1,198 @@
...
+                retStr += '\n'
+            }
+            i++
+        }
+
+        // pad to minimum length
+        while(retStr.length < 'BR, EUNE, EUW, JP,'.length) {
+            retStr += " "
+        }
+        return retStr
```
molenzwiebel: There are a bunch of missing semi-colons here. Also, `Array.prototype.join` is a thing.

src/ApiStatus.ts
```diff
@@ -0,0 +1,198 @@
...
+                statusEmbed.api[api] = retStr;
+        }
+
+        statusEmbed["onFire"] = onFire || (apiIssuesCounter / apiCounter) > this.sharedSettings.apiStatus.apiOnFireThreshold;
+        statusEmbed["allApisOK"] = allApisOK;
+        statusEmbed["allApisIssues"] = allApisIssues;
+        return statusEmbed;
+    }
+
+    private joinArray(arr: Array<string>): string {
```
molenzwiebel: `string[]`

src/ApiStatus.ts
```diff
@@ -0,0 +1,198 @@
...
+                retStr += ":x:" + this.joinArray(regionStates.down) + "\n";
+            }
+            if (regionStates.up.length > 0) {
+                allApisIssues = false;
+            }
+            apiIssuesCounter += regionStates.troubled.length + regionStates.down.length;
+            apiCounter += regionStates.troubled.length + regionStates.down.length + regionStates.up.length;
+
+            // API on fire, if all regions for one api have issues
+            if (regionStates.troubled.length + regionStates.down.length == regionCounter) {
```
molenzwiebel: `===`

src/ApiStatus.ts
```diff
@@ -0,0 +1,198 @@
...
+        let cacheObject: { [key: string]: any } = {};
+
+        let onFire = false;
+        let allApisOK = true;
+        let allApisIssues = false;
+        let apiIssuesCounter = 0;
+        let apiCounter = 0;
+        let statusEmbed: StatusEmbedState = { api: {}, onFire: false, allApisOK: true, allApisIssues: false};
+
+        type RegionState = {"troubled": Array<string>, "up": Array<string>, "down": Array<string>};
```
molenzwiebel: `string[]`

src/ApiStatus.ts
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+        message.channel.send({embed: embedContent});
+    }
+
+    private async getApiStatus(): Promise<StatusEmbedState> {
+        // cache embed state
+        let timeDiff = Date.now() - this.lastCheckTime;
+        if (timeDiff > this.sharedSettings.apiStatus.checkInterval) {
+            let apiStatus = await this.apiStatusAPI.getApiStatus();
```
molenzwiebel: `const`

src/ApiStatus.ts
```diff
@@ -0,0 +1,198 @@
...
+        if (apiStatus.onFire) {
+            embedContent.image = { url: this.pickRandomOnFireImage() };
+        }
+
+        message.channel.send({embed: embedContent});
+    }
+
+    private async getApiStatus(): Promise<StatusEmbedState> {
+        // cache embed state
+        let timeDiff = Date.now() - this.lastCheckTime;
```
molenzwiebel: `const`

src/ApiStatus.ts
```diff
@@ -0,0 +1,198 @@
...
+            },
+            fields: fields
+        }
+
+        if (apiStatus.onFire) {
+            embedContent.image = { url: this.pickRandomOnFireImage() };
+        }
+
+        message.channel.send({embed: embedContent});
+    }
```
molenzwiebel: Bunch of missing semi-colons and `let`s instead of `const`s in this function too

src/ApiStatus.ts
```diff
@@ -0,0 +1,198 @@
...
+    onFire: boolean;
+    allApisOK: boolean;
+    allApisIssues: boolean;
+}
+
+export default class ApiStatus {
+    private bot: Discord.Client;
+    private sharedSettings: SharedSettings;
+    private command: string;
+    private apiStatusAPI: APIStatusAPI
```
molenzwiebel: Semi-colon ;)

src/ApiStatus.ts
TiFu: I don't think Array.prototype.join would help me much in this case. I only add a `\n` every 4 elements. This is not possible with `join` (unless I use `Array.slice` on 4 elements each, which would make the code less readable imo).

src/ApiStatus.ts
```diff
@@ -50,19 +49,19 @@ export default class ApiStatus {
         if (!this.isApiStatusCommand(content))
             return;
 
-        let apiStatus = await this.getApiStatus();
+        const apiStatus = await this.getApiStatus();
 
-        let fields: Array<{name: string, value: string, inline: boolean}> = [];
-        for (let api in apiStatus.api){ 
-            fields.push({name: api, value: apiStatus.api[api], inline: true})
+        const fields: Array<{name: string, value: string, inline: boolean}> = [];
```
molenzwiebel: Minor nitpick, but this should technically also be `{ name: string, value: string, inline: boolean }[]`. You should be able to omit the type entirely though, the latest TypeScript versions infer array types based on what kind of elements you put in them (although it might not infer this properly since it isn't guaranteed to be filled).

src/ApiStatus.ts
```diff
@@ -91,11 +90,23 @@ export default class ApiStatus {
...
+        const min = Math.floor(timeDiff / 1000 / 60);
+        if (min >= 1) {
+            const minutes = min == 1 ? "minute" : "minutes";
+            return "Last refresh: " + min + " " + minutes +" ago";
+        } else {
+            const sec = Math.floor((timeDiff - min * 60000) / 1000);
+            const seconds = sec == 1 ? "second" : "seconds";
+            return "Last refresh: " + sec + " " + seconds + " ago";
+        }
+    }
```
molenzwiebel: We already depend on [pretty-ms](https://github.com/sindresorhus/pretty-ms) for the uptime command, so this can probably be simplified to `prettyMs(Date.now() - this.lastCheckTime, { verbose: true }) + " ago"` (with the appropriate import of course).

src/ApiStatus.ts
```diff
@@ -104,22 +115,22 @@ export default class ApiStatus {
...
         let onFire = false;
         let allApisOK = true;
         let allApisIssues = false;
         let apiIssuesCounter = 0;
         let apiCounter = 0;
-        let statusEmbed: StatusEmbedState = { api: {}, onFire: false, allApisOK: true, allApisIssues: false};
+        const statusEmbed: StatusEmbedState = { api: {}, onFire: false, allApisOK: true, allApisIssues: false};
 
-        type RegionState = {"troubled": Array<string>, "up": Array<string>, "down": Array<string>};
+        type RegionState = {"troubled": string[], "up": string[], "down": string[]};
```
molenzwiebel: It might be better to move this to the top level with the other type declarations. You can also remove the quotes from the key names, with the same rationale as in the previous review.

src/ApiStatus.ts
TiFu: It infers as `any[]`. I'll change it to `{name: string, ...}[]`

src/ApiStatus.ts
TiFu: I will remove the quotes and move it into line 130 (which is the only place where this is used).

src/ApiStatus.ts
```diff
@@ -92,16 +93,9 @@ export default class ApiStatus {
...
-        const min = Math.floor(timeDiff / 1000 / 60);
-        if (min >= 1) {
-            const minutes = min == 1 ? "minute" : "minutes";
-            return "Last refresh: " + min + " " + minutes +" ago";
-        } else {
-            const sec = Math.floor((timeDiff - min * 60000) / 1000);
-            const seconds = sec == 1 ? "second" : "seconds";
-            return "Last refresh: " + sec + " " + seconds + " ago";
-        }
+        return "Last refresh: " + prettyMs(Math.max(timeDiff, 1000), { verbose: true, secDecimalDigits: 0 }) + " ago";
```
Querijn: Is this correct? Specifically the part where it cant be lower than 1 second.

TiFu: I did that on purpose. I wanted to avoid showing `5 milliseconds ago` to the user. That adds as much information as 1 second ago. 
The update timer is also not entirely correct. It would be better if your api status api returns a 'lastUpdateTime' field and we use that to show the duration since the last update.

PR #16, opened 2018-03-06
Add option to not receive reactions 

stelar7:


Querijn:
Looks good, can you edit the welcome message to indicate they can opt out of reacts?

src/AutoReact.ts
```diff
@@ -66,18 +88,23 @@ export default class AutoReact {
...
-            && !greeting.startsWith("hey ") && greeting != "hey"
-            && !greeting.startsWith("good morning ") && greeting != "good morning"
-            && !greeting.startsWith("goodmorning ") && greeting != "goodmorning"
-            && !greeting.startsWith("good evening ") && greeting != "good evening"
-            && !greeting.startsWith("goodevening ") && greeting != "goodevening"
-            && !greeting.startsWith("good night ") && greeting != "good night"
-            && !greeting.startsWith("goodnight ") && greeting != "goodnight"
-            && !greeting.startsWith("goodday ") && greeting != "goodday"
-            && !greeting.startsWith("good day ") && greeting != "good day")
+        const words: String[] = [
```
molenzwiebel: `string`, not `String` ;)

src/AutoReact.ts
```diff
@@ -66,18 +88,23 @@ export default class AutoReact {
...
+            "hello", "hi", "hey",
+            "good morning", "goodmorning",
+            "good evening", "goodevening",
+            "good night", "goodnight",
+            "good day", "goodday"
+        ];
+
+        const shouldReact = words.filter((x: string) => !greeting.startsWith(x)).filter((x: string) => greeting != x).length != words.length;
```
molenzwiebel: You should be able to leave out the casts here if you correct the previous comments. Also, if I'm reading this correctly, it should be able to be simplified to just `words.some(x => greeting.startsWith(x))`, which looks a lot cleaner.

src/AutoReact.ts
```diff
@@ -66,18 +88,23 @@ export default class AutoReact {
...
-            && !greeting.startsWith("hey ") && greeting != "hey"
-            && !greeting.startsWith("good morning ") && greeting != "good morning"
-            && !greeting.startsWith("goodmorning ") && greeting != "goodmorning"
-            && !greeting.startsWith("good evening ") && greeting != "good evening"
-            && !greeting.startsWith("goodevening ") && greeting != "goodevening"
-            && !greeting.startsWith("good night ") && greeting != "good night"
-            && !greeting.startsWith("goodnight ") && greeting != "goodnight"
-            && !greeting.startsWith("goodday ") && greeting != "goodday"
-            && !greeting.startsWith("good day ") && greeting != "good day")
+        const words: string[] = [
```
molenzwiebel: You shouldn't even need the type specifier here, but I'll allow it 😛

Querijn:
Original thinking was broken, seems like it got some toggle code of its own, but the reply of the bot was wrong.

PR #26, opened 2018-03-13
Add command manager

### LINKED ISSUE #24: Switch to command manager
I opted out of this feature because it's pretty complex, and I figured we'd only have 2 or 3 commands, but since it's racking up new ones really quickly, we might wanna go for something that manages these commands. It means that we could output help for all of those as well. 

Whoever takes this on, there's a command manager here for our language here: https://github.com/discordjs/Commando

stelar7:
fixes #24 (except it doesnt use that lib...)

### DIFF
--- a/settings/shared_settings.json
+++ b/settings/shared_settings.json
@@ -1,7 +1,8 @@
 {
     "server": "342988445498474498",
     "botty": {
-        "prefix": "!"
+        "prefix": "!",
+        "nickname":"Botty McBotface"
     },
     "channelAccess": {
         "forcedChannels": [
--- a/settings/shared_settings.json
+++ b/settings/shared_settings.json
@@ -2,7 +2,7 @@
     "server": "342988445498474498",
     "botty": {
         "prefix": "!",
-        "nickname":"Botty McBotface"
+        "nickname": "Botty McBotface"
     },
     "channelAccess": {
         "forcedChannels": [
@@ -13,15 +13,30 @@
         "restrictedChannels": [
             "190817288507752448"
         ],
-        "commands":[
-            "join",
-            "leave"
+        "commands": [
+            {
+                "aliases": [
+                    "join"
+                ],
+                "description": "Joins a channel"
+            },
+            {
+                "aliases": [
+                    "leave"
+                ],
+                "description": "Leaves a channel"
+            }
         ]
     },
     "uptimeSettings": {
         "checkInterval": 60000,
         "commands": [
-            "uptime"
+            {
+                "aliases": [
+                    "uptime"
+                ],
+                "description": "Checks the uptime of botty"
+            }
         ]
     },
     "techBlog": {
@@ -47,7 +62,14 @@
             "350692046631206912",
             "347539422943903755"
         ],
-        "command": "note"
+        "commands": [
+            {
+                "aliases": [
+                    "note"
+                ],
+                "description": "this hasnt been changed to use the command system yet..."
+            }
+        ]
     },
     "officehours": {
         "allowedRoles": [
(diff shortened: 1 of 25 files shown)


settings/shared_settings.json
```diff
@@ -45,24 +55,37 @@
...
-        "baseURL":"https://api.github.com/repos/WxWatch/riot-api-libraries/contents/libraries/"
+        "baseURL": "https://api.github.com/repos/WxWatch/riot-api-libraries/contents/libraries/"
     },
     "autoReact": {
-        "emoji": "408527155891273738"
+        "emoji": "408527155891273738",
+        "aliases":[
+            "toggle_default_thinking",
+            "toggle_react",
+            "refresh_thinking"
```
Querijn: not very descriptive

also it stops reacting to hi

src/OfficeHours.ts
```diff
@@ -79,63 +79,62 @@ export default class OfficeHours {
         }
     }
 
-    onCommand(message: Discord.Message) {
+    onCommand(sender: Discord.User, channel: Discord.TextChannel, message: Discord.Message, command: string, args: string[]) {
```
Querijn: sender and channel are in message, why are they separated?

src/Botty.ts
```diff
@@ -88,4 +89,32 @@ export default class Botty {
     start() {
         return this.client.login(this.personalSettings.discord.key);
     }
+
+    private commands: CommandBase[] = [];
```
molenzwiebel: Move this up with the other instance variables.

src/Botty.ts
```diff
@@ -88,4 +89,32 @@ export default class Botty {
...
+
+        if (message.author.bot) return;
+        if (!message.content.startsWith(this.sharedSettings.botty.prefix)) return;
+
+        const parts = message.content.split(" ");
+        const command = parts[0].substr(1);
+
+        this.commands.forEach(cmd => {
+            if (cmd.alias === command) {
+                cmd.handler.onCommand(message.author, <Discord.TextChannel>message.channel, message, command, parts.slice(1))
```
molenzwiebel: Missing `;`

src/OfficeHours.ts
```diff
@@ -19,29 +20,28 @@ interface OfficeHoursData {
...
-        this.bot.on("ready", this.onBot.bind(this));
-        this.bot.on("message", this.onCommand.bind(this));
-		// this.bot.on("channelUpdate", this.onChannelUpdate.bind(this));
+        // this.bot.on("channelUpdate", this.onChannelUpdate.bind(this));
     }
 
-    onBot() {
-        const mainGuild = this.bot.guilds.get(this.sharedSettings.server);
+    onReady(bot: Discord.Client) {
+        this.bot = bot;
```
molenzwiebel: Any cleaner way to do this? This seems a bit hacky.

src/OfficeHours.ts
```diff
@@ -151,28 +150,28 @@ export default class OfficeHours {
...
-        .then(event => { 
-            channel.send(`On this day in ${event.year}, ${event.text}`);
-        })
-        .catch(r => console.warn('Failed to fetch message of the day: ' + r));
+            .then(r => r.json())
+            .then(({ data: { Events } }) => Events[Math.floor(Math.random() * Events.length)])
+            .then(event => {
+                channel.send(`On this day in ${event.year}, ${event.text}`);
+            })
+            .catch(r => console.warn('Failed to fetch message of the day: ' + r));
```
molenzwiebel: Might be cleaner to convert this to an `async` method now that we're here anyway 😉

src/OfficeHours.ts
```diff
@@ -192,40 +191,40 @@ export default class OfficeHours {
...
+                    if (usersToMention.length > 0)
+                        channel.send(usersToMention.join(", ") + "\n");
+                }
+
+                this.sendOnThisDayMessage(channel);
+            })
+            .catch(reason => {
+                console.warn("Failed getting last close message: " + reason);
+                this.sendOnThisDayMessage(channel);
+            });
```
molenzwiebel: This can also be async (the method is already async lul)

src/RiotAPILibraries.ts
```diff
@@ -71,58 +67,31 @@ export default class RiotAPILibraries {
...
-            }
+        return this.getListForLanguage(channel, param);
+    }
 
-            let editMessage = await editMessagePromise; 
-            if (Array.isArray(editMessage)) editMessage = editMessage[0];
-            editMessage.edit({ embed });
-        }
+    invalidArguments(channel: Discord.TextChannel, arg: string[]) {
+        channel.send(`unknown argument for command; ${arg}`);
```
molenzwiebel: This can be a lot cleaner than this ;)

src/RiotAPILibraries.ts
```diff
@@ -150,6 +119,47 @@ export default class RiotAPILibraries {
...
+            return;
+        }
+
+        const libraryList = await response.json();
+        if (!Array.isArray(libraryList) || libraryList.length === 0 || !libraryList[0].sha) {
+            channel.send(this.settings.riotApiLibraries.noLanguage + language);
+            return;
+        }
+
+        let editMessagePromise = channel.send(`Found the list of libraries for ${language}, listing ${libraryList.length} libraries, this post will be edited with the result.`);
```
molenzwiebel: `const`

settings/shared_settings.json
stelar7: I haven't touched the react code, so that should still work.. :thinking:

src/OfficeHours.ts
stelar7: Its either gonna end up in onReady or the constructor, and I'd prefer having the constructors as clean as possible :man_shrugging: 

(Tho I managed to remove it altogether from officehours, it still remains in some of the other classes)

settings/shared_settings.json
Querijn: No what I mean is that the "thinking" isn't the only reaction it has, it also stops the triggers for greetings

Querijn:
Gonna put this one on hold. The issue I currently have is that the command's intention is completely hidden.

Can you tell what this function does?
afbeelding

Usually the startsWith("!libs") gave it away immediately, but since that line is now gone, it's hard to tell which is which, especially since the calls in app.ts don't give it away either. It took me a while to find out which one actually was used for help.

I have this setup in the "stelar7-commands" branch, which I'm going to keep out of main branch for two reasons, the first being that !help is currently in use by Heimerdinger, which will go down as soon as it's merged with Botty, and the second being issues described above. I'll merge it if I find a good way to more easily define which function does what exactly.

PR #28, opened 2018-03-17
Add similar-note feature

### LINKED ISSUE #25: Note suggestions and aliases
If you don't get the `!note` call right, we should show a suggestion like "Hey, did you mean .ddragon?"

I'm not sure how to approach this. Could use some expertise :)

stelar7:
fixes #25

### DIFF
--- a/src/Info.ts
+++ b/src/Info.ts
@@ -168,8 +168,6 @@ export default class Info {
         });
 
         if (!info) {
-            let message = "Did you mean: ";
-
             const data = this.infos.slice()
                 .map(i => {
                     return {
@@ -179,6 +177,7 @@ export default class Info {
                 }).filter(s => s.score < 2);
 
             if (data.length !== 0) {
+                let message = "Did you mean: ";
                 message += data.map(s => "`" + s.command + "`").join(", ") + "?";
                 return { message, counter: 0, command };
             }
--- a/src/Info.ts
+++ b/src/Info.ts
@@ -162,11 +162,34 @@ export default class Info {
     }
 
     private fetchInfo(command: string): InfoData | null {
+
         const info = this.infos.find(inf => {
             return inf.command === command;
         });
 
-        if (!info) { return null; }
+        if (!info) {
+            const similarity = this.infos.slice().map(i => {
+                return {
+                    command: i.command,
+                    score: this.levenshteinDistance(command, i.command),
+                };
+            });
+
+            let message = "Did you mean: ";
+            let found = false;
+            similarity.forEach(s => {
+                if (s.score < 2) {
+                    message += "`" + s.command + "`, ";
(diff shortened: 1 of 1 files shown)


Querijn:
nice

src/Info.ts
```diff
@@ -162,11 +162,34 @@ export default class Info {
...
+            });
+
+            let message = "Did you mean: ";
+            let found = false;
+            similarity.forEach(s => {
+                if (s.score < 2) {
+                    message += "`" + s.command + "`, ";
+                    found = true;
+                }
+            });
```
Querijn: This could be a map() and join()'d, would save that last slice

src/Info.ts
```diff
@@ -176,4 +199,40 @@ export default class Info {
...
+        if (a.length === 0) {
+            return b.length;
+        }
+
+        if (b.length === 0) {
+            return a.length;
+        }
+
+        let v0 = [];
+        const v1 = [];
```
Querijn: one const, one let?

stelar7: v0 is changed later in the loop, so it cant be const.

src/Info.ts
```diff
@@ -162,11 +162,34 @@ export default class Info {
...
+            let found = false;
+            similarity.forEach(s => {
+                if (s.score < 2) {
+                    message += "`" + s.command + "`, ";
+                    found = true;
+                }
+            });
+
+            if (found) {
+                return { message: message.slice(0, -2) + "?", counter: 0 } as InfoData;
```
molenzwiebel: Do you really need to cast here?

src/Info.ts
```diff
@@ -162,11 +162,34 @@ export default class Info {
...
+            }
+
+            return { message: `No note with the name ${command} was found.`, counter: 0 } as InfoData;
```
molenzwiebel: ^

src/Info.ts
```diff
@@ -176,4 +199,40 @@ export default class Info {
         info.counter++;
         return info;
     }
+
+    private levenshteinDistance(a: string, b: string): number {
```
molenzwiebel: Cmon stelar, this is not the node mentality. Get yourself a package instead! [fast-levenshtein](https://www.npmjs.com/package/fast-levenshtein) has typescript typings available at `@types/fast-levenshtein`.

src/Info.ts
stelar7: Yeah

src/Info.ts
stelar7: ^

src/Info.ts
Querijn: If you require a cast, it means something isn't set. In this case, you're missing "command"

PR #29, opened 2018-03-17
print note-list if no params are passed

### LINKED ISSUE #22: !note/notes should also list all notes
To avoid confusion.

stelar7:
Fixes #22


stelar7:
This is also fixed in #26, so if you merge that you can close this.

Issue #30, opened 2018-03-17
Missing rejection handling for a promise


Querijn:
> Error: You have triggered an unhandledRejection, you may have forgotten to catch a Promise rejection:
> TypeError: Cannot read property 'options' of undefined
>     at item.request.gen.end (/home/pi/bot/node_modules/discord.js/src/client/rest/RequestHandlers/Sequential.js:69:74)
>     at then (/home/pi/bot/node_modules/snekfetch/src/index.js:218:21)
>     at process._tickDomainCallback (internal/process/next_tick.js:135:7)

Every once in a while we get the nice error message described above.. We need to find it and handle it correctly. It doesn't break anything as far as I know, but let's handle it before it shuts down Botty every time it happens.

Querijn:
This is a Discord.js bug

PR #33, opened 2018-03-26
add try/catch to all awaits, replace .then with await

stelar7:
possible fix for #30

### DIFF
--- a/src/AnswerHub.ts
+++ b/src/AnswerHub.ts
@@ -75,20 +75,24 @@ export default class AnswerHubAPI {
      * @returns The parsed body of the response from the AnswerHubAPI
      */
     private async makeRequest<T>(url: string): Promise<T> {
-        const resp = await fetch(`${this.baseURL}services/v2/${url}`, {
-            headers: {
-                "Accept": "application/json",
-                "Authorization": this.auth,
-                "Content-Type": "application/json",
-            },
-            method: "POST",
-        });
+        try {
+            const resp = await fetch(`${this.baseURL}services/v2/${url}`, {
+                headers: {
+                    "Accept": "application/json",
+                    "Authorization": this.auth,
+                    "Content-Type": "application/json",
+                },
+                method: "POST",
+            });
 
-        if (resp.status !== 200) {
-            throw new Error(`Received status code ${resp.status}`);
+            if (resp.status !== 200) {
+                throw new Error(`Received status code ${resp.status}`);
+            }
+
+            return resp.json();
+        } catch (error) {
+            console.error(`Error occurred while making a request to the answerhub api: ${error}`);
         }
-
-        return resp.json();
     }
 }
 
--- a/src/ApiStatus.ts
+++ b/src/ApiStatus.ts
@@ -53,7 +53,15 @@ export default class ApiStatus {
         const content = message.cleanContent;
         if (!this.isApiStatusCommand(content)) return;
 
-        const apiStatus = await this.getApiStatus();
+        let apiStatus: StatusEmbedState;
+        try {
+            apiStatus = await this.getApiStatus();
+        } catch (error) {
+            console.error(error);
+            // fake status so we dont need to wrap the entire block
+            apiStatus = { api: {}, onFire: true, allApisOK: false, allApisIssues: true };
+            console.error(`Error occurred while making a request to the apistatus api: ${error}`);
+        }
 
         const fields: { name: string, value: string, inline: boolean }[] = [];
         for (const api in apiStatus.api) {
@@ -100,15 +108,20 @@ export default class ApiStatus {
     }
 
     private async getApiStatus(): Promise<StatusEmbedState> {
-        // cache embed state
-        const timeDiff = Date.now() - this.lastCheckTime;
-        if (timeDiff > this.sharedSettings.apiStatus.checkInterval) {
-            const apiStatus = await this.apiStatusAPI.getApiStatus();
-            this.currentStatus = this.parseApiStatus(apiStatus);
-            this.lastCheckTime = Date.now();
-        }
+        try {
+            // cache embed state
+            const timeDiff = Date.now() - this.lastCheckTime;
+            if (timeDiff > this.sharedSettings.apiStatus.checkInterval) {
+                const apiStatus = await this.apiStatusAPI.getApiStatus();
+                this.currentStatus = this.parseApiStatus(apiStatus);
+                this.lastCheckTime = Date.now();
+            }
+            return this.currentStatus;
 
-        return this.currentStatus;
+        } catch (error) {
+            console.error(`Error occurred while making a request to the apistatus api: ${error}`);
+            return this.currentStatus;
+        }
     }
 
     private parseApiStatus(apiStatus: APIStatus): StatusEmbedState {
--- a/src/ApiStatus.ts
+++ b/src/ApiStatus.ts
@@ -53,15 +53,7 @@ export default class ApiStatus {
         const content = message.cleanContent;
         if (!this.isApiStatusCommand(content)) return;
 
-        let apiStatus: StatusEmbedState;
-        try {
-            apiStatus = await this.getApiStatus();
-        } catch (error) {
-            console.error(error);
-            // fake status so we dont need to wrap the entire block
-            apiStatus = { api: {}, onFire: true, allApisOK: false, allApisIssues: true };
-            console.error(`Error occurred while making a request to the apistatus api: ${error}`);
-        }
+        const apiStatus = await this.getApiStatus();
 
         const fields: { name: string, value: string, inline: boolean }[] = [];
         for (const api in apiStatus.api) {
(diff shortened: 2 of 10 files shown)


src/AnswerHub.ts
```diff
@@ -75,20 +75,24 @@ export default class AnswerHubAPI {
...
+
+            if (resp.status !== 200) {
+                throw new Error(`Received status code ${resp.status}`);
+            }
 
-        if (resp.status !== 200) {
-            throw new Error(`Received status code ${resp.status}`);
+            return resp.json();
+        } catch (error) {
+            console.error(`Error occurred while making a request to the answerhub api: ${error}`);
```
molenzwiebel: This method will now resolve with `undefined` instead of reject whenever an error is encountered. Is that intended, or did you forget to rethrow here?

src/ApiStatus.ts
```diff
@@ -53,7 +53,15 @@ export default class ApiStatus {
...
 
-        const apiStatus = await this.getApiStatus();
+        let apiStatus: StatusEmbedState;
+        try {
+            apiStatus = await this.getApiStatus();
+        } catch (error) {
+            console.error(error);
+            // fake status so we dont need to wrap the entire block
+            apiStatus = { api: {}, onFire: true, allApisOK: false, allApisIssues: true };
+            console.error(`Error occurred while making a request to the apistatus api: ${error}`);
```
molenzwiebel: Printing the error twice 🤔

molenzwiebel: Also, this will never get called since there's no way `getApiStatus` will reject (see comments below).

src/ApiStatus.ts
```diff
@@ -100,15 +108,20 @@ export default class ApiStatus {
...
+                const apiStatus = await this.apiStatusAPI.getApiStatus();
+                this.currentStatus = this.parseApiStatus(apiStatus);
+                this.lastCheckTime = Date.now();
+            }
+            return this.currentStatus;
 
-        return this.currentStatus;
+        } catch (error) {
+            console.error(`Error occurred while making a request to the apistatus api: ${error}`);
+            return this.currentStatus;
```
molenzwiebel: Due to the catch here this method will never reject. Which means that the handlers above will never reach the error case ;)

molenzwiebel: Also, bit unrelated, but it looks like both this class and the api status API cache the result. Maybe we can remove the caching here?

src/ApiStatusApi.ts
```diff
@@ -28,24 +28,29 @@ export default class ApiStatusApi {
...
+            });
+
+            if (resp.status !== 200) {
+                throw new Error(`[ApiStatus] Received status code ${resp.status}`);
+            }
+
+            this.cached = await resp.json();
+            this.lastUpdate = Date.now();
+        } catch (error) {
+            console.error(`Error occurred while making a request to the apistatus api: ${error}`);
```
molenzwiebel: Again, you're swallowing errors here that previously would have caused the promise to reject. Rethrow here?

src/ForumReader.ts
```diff
@@ -82,13 +82,17 @@ export default class ForumReader {
...
         } else {
-            const node = await this.answerHub.getNode(id);
-            this.cachedNodes.set(id, node);
-            return node;
+            try {
+                const node = await this.answerHub.getNode(id);
+                this.cachedNodes.set(id, node);
+                return node;
+            } catch (error) {
+                console.error(`Error occurred while making a request to the answerhub api: ${error}`);
```
molenzwiebel: So while this is a valid way to indicate errors, it's a bit "unconventional". I'm fine with both options (resolving as undefined or rejecting), but be sure to update the documentation since right now its incorrect.

src/ForumReader.ts
```diff
@@ -160,7 +164,11 @@ export default class ForumReader {
         this.keyFinder.findKey(username, activity.body, embed.url as string, activity.creationDate);
         embed.setTimestamp(new Date(activity.creationDate)).setThumbnail(avatar);
 
-        await this.channel.send({ embed });
+        try {
+            await this.channel.send({ embed });
+        } catch (error) {
+            console.error(`Error occurred while making a request to the answerhub api: ${error}`);
```
molenzwiebel: Nice copy and paste

src/KeyFinder.ts
```diff
@@ -43,21 +43,26 @@ export default class KeyFinder {
...
     public async findKey(user: string, message: string, location: string, timestamp: number): Promise<boolean> {
         const matches = message.match(/(RGAPI-)?[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}/ig);
         if (!matches) { return false; }
 
         let found = false;
         for (const match of matches) {
-            const limit = await this.testKey(match);
+            let limit;
+            try {
+                limit = this.testKey(match);
```
molenzwiebel: You forgot an `await` here ;)

src/KeyFinder.ts
```diff
@@ -43,21 +43,26 @@ export default class KeyFinder {
...
+            } catch (error) {
+                console.error(`Error occurred while making a request to the riot games api: ${error}`);
```
molenzwiebel: This will never get called since `testKey` never rejects.

src/KeyFinder.ts
```diff
@@ -116,32 +121,41 @@ export default class KeyFinder {
...
             const keyInfo = this.keys[i];
-            this.testKey(keyInfo.apiKey).then(header => {
+
+            try {
+                const header = await this.testKey(keyInfo.apiKey);
                 if (header !== null) {
                     return;
                 }
+            } catch (error) {
+                console.error(`Error occurred while making a request to the riot games api: ${error}`);
```
molenzwiebel: If I'm reading this correctly it will remove keys that error from the database, even though they might still be valid. Is that intentional?

molenzwiebel: Also, just like the other place where this is called, `testKey` will never reject so this will never get called.

src/OfficeHours.ts
```diff
@@ -188,42 +215,46 @@ export default class OfficeHours {
...
-            });
+            // Find all users that raised their hand
+            const reactions = closeMessage.reactions.get("✋");
+            if (reactions) {
+                const usersToMention = reactions.users.array().filter(user => !user.bot);
+                if (usersToMention.length > 0) {
+                    channel.send(usersToMention.join(", ") + "\n");
+                }
+            }
+            this.sendOnThisDayMessage(channel);
```
molenzwiebel: Do you want an `await` here?

src/OfficeHours.ts
```diff
@@ -188,42 +215,46 @@ export default class OfficeHours {
...
+                message = message[0];
+            }
 
-                this.data.lastCloseMessage = message.id;
-                message.react("✋");
-            });
+            this.data.lastCloseMessage = message.id;
+            message.react("✋");
+        } catch (error) {
+            console.error(`Error occurred while sending close message: ${error}`);
```
molenzwiebel: You might want to set `lastCloseMessage` to null here so the next time office hours reopens it won't operate on a stale message.

src/RiotAPILibraries.ts
```diff
@@ -84,76 +84,95 @@ export default class RiotAPILibraries {
...
-            }
+                try {
+                    const responses = await fetch(this.settings.githubLibraries.baseURL, this.fetchSettings);
+                    const data = await responses.json() as GithubAPIStruct[];
+
+                    const languages = "`" + data.map(x => x.name).join(", ") + "`";
+                    const reply = this.settings.githubLibraries.languageList.replace("{languages}", languages);
+                    message.channel.send(reply);
+                } catch (error) {
+                    console.error(`Error occurred while fetching from github: ${error}`);
```
molenzwiebel: Maybe send a message to the user, so they aren't left waiting for nothing?

src/RiotAPILibraries.ts
```diff
@@ -84,76 +84,95 @@ export default class RiotAPILibraries {
...
-            }
 
-            let editMessage = await editMessagePromise;
-            if (Array.isArray(editMessage)) { editMessage = editMessage[0]; }
-            editMessage.edit({ embed });
+                let editMessage = await editMessagePromise;
+                if (Array.isArray(editMessage)) { editMessage = editMessage[0]; }
+                editMessage.edit({ embed });
+            } catch (error) {
+                console.error(`Error occurred fetching from github: ${error}`);
```
molenzwiebel: Send something to the user? There might be a stale message where botty promises to edit something

src/RiotAPILibraries.ts
```diff
@@ -84,76 +84,95 @@ export default class RiotAPILibraries {
...
+                valid: true,
+            };
+        } catch (error) {
+            console.error(`Error occurred while fetching from github: ${error}`);
+
+            return {
+                library: null,
+                links: [],
+                stars: -1,
+                valid: true,
```
molenzwiebel: valid: true? 🤔

Querijn:
Wrapping everything in a try/catch will not fix the issue. I'd rather be punched by botty with a clear tracelog than Botty silently restarting 4 times a minute.

stelar7:
Wrapping everything in a trycatch will handle all promise rejection, wont it? :thinking:

src/OfficeHours.ts
stelar7: Fixed, github wont show as outdated for some reason..

src/RiotAPILibraries.ts
stelar7: Fixed, github wont show as outdated for some reason..

src/RiotAPILibraries.ts
stelar7: Fixed, github wont show as outdated for some reason..

Querijn:
This doesn't fix #30 because it happens in the request queue of Discord.js. You can't catch that one, the library would have to do it.

The fix would be whatever this is:
https://github.com/discordjs/discord.js/issues/1903

PR #35, opened 2018-03-26
sort by replacement distance

stelar7:


src/Info.ts
```diff
@@ -181,7 +181,9 @@ export default class Info {
                         command: i.command,
                         score: this.levenshteinDistance(command, i.command),
                     };
-                }).filter(s => s.score < this.sharedSettings.info.maxScore);
+                })
+                .filter(s => s.score < this.sharedSettings.info.maxScore)
+                .sort((a, b) => a.score > b.score ? 1 : -1);
```
molenzwiebel: a.score - b.score

PR #44, opened 2018-03-27
Updated declarations and If statement cleanliness

Sir-Kam:
changed -includes( . . . )- to -indexOf(...)-

### DIFF
--- a/src/Info.ts
+++ b/src/Info.ts
@@ -54,11 +54,11 @@ export default class Info {
         // if using .syntax we can only read notes
         let commandIsFetch = false;
 
-        // Needs to start with '/' or '!' or in separate cases '.'
-        const split = message.cleanContent.split(/[\n\r\s]/);
-        if (split[0][0] === ".") {
-            commandIsFetch = true;
-        } else if (split[0][0] !== "!" && split[0][0] !== "/") return;
+        // Needs to start with '/' / '!' or '.'
+        const contentPrefix = message.cleanContent[0];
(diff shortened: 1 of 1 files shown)


src/Info.ts
```diff
@@ -54,11 +54,11 @@ export default class Info {
         // if using .syntax we can only read notes
         let commandIsFetch = false;
```
stelar7: this can be removed and it can be made const where you assign it

src/Info.ts
```diff
@@ -54,11 +54,11 @@ export default class Info {
...
 
-        // Needs to start with '/' or '!' or in separate cases '.'
+        // Needs to start with '/' / '!' or '.'
+        const contentPrefix = message.cleanContent[0];
```
stelar7: split[0][0]?

Sir-Kam: This gets the first char in the string. So the use of that was unnecessary. But either could be used, just in my opinion, i thought it looked messy; Though, I am re-referencing the cleanContent, when the same data is in 'split'.

Querijn:
Thanks for your commit. Unfortunately I've taken a good hard look at Info and I see that things could be done far better. With the changes I made to include '.' as a prefix for fetch I should've split up this prefix into its own handler rather than in the same function as the !note handler. 

I believe that's the way forward for the note functionality. Next to that, it could use some other changes, making it dependent on the commands a lot for cleanliness.

Sir-Kam:
Agreed.

PR #52, opened 2018-03-27
fix lint issues

stelar7:


stelar7:
somehow the previous lint merge got overwritten..?

Querijn:
Assuming no change in functionality, went through it, one place where var `message` was changed to `response`, but I assume that's fine

stelar7:
Yeah, the var was shadowing the function parameter, so i changed it

Issue #48, opened 2018-03-27
Finish command manager.


Querijn:
- [ ] Make commands toggable.
- [ ] Separate commands per handler.
- [ ] Test

stelar7:
I've got this working on the [command branch](https://github.com/stelar7/BottyMcBotface/tree/commands) now, with the caveat that all command functions must be called with the arrow syntax as seen [here](https://github.com/stelar7/BottyMcBotface/blob/commands/src/CommandController.ts#L72), not sure if you still wanna do it yourself or pull the branch for tests?

stelar7:
This can be closed?

PR #54, opened 2018-03-29
Command handler v2

stelar7:
updates to the stelar7-commands branch, checks 1. and 2. of #48


src/OfficeHours.ts
```diff
@@ -215,16 +211,15 @@ export default class OfficeHours {
         const everyone = channel.guild.roles.find("name", "@everyone");
         await channel.overwritePermissions(everyone, { SEND_MESSAGES: false });
 
-        channel.send(this.sharedSettings.officehours.closeMessage.replace(/{botty}/g, this.bot.user.toString()))
-            .then(message => {
-                if (Array.isArray(message)) {
-                    message = message[0];
-                }
+        let message = await channel.send(this.sharedSettings.officehours.closeMessage.replace(/{botty}/g, this.sharedSettings.botty.nickname));
```
Querijn: Replace it with this.bot.user, doesn't require a setting.

settings/shared_settings.json
```diff
@@ -1,13 +1,15 @@
...
-        "forcedChannels": [
-            "301148922443726870",
-            "187961895049756672",
-            "190817288507752448"
-        ],
-        "restrictedChannels": [
-            "190817288507752448"
+    "botty": {
+        "prefix": "!",
+        "nickname": "Botty McBotface"
```
Querijn: can be removed as far as I can tell

src/app.ts
```diff
@@ -13,28 +14,55 @@ import Techblog from "./Techblog";
...
+// register commands
+controller.registerCommand(commandList.controller.toggle, controller.onToggle);
+controller.registerCommand(commandList.controller.help, controller.onHelp);
+
+const notes = new Info(sharedSettings, "data/info_data.json", versionChecker);
+controller.registerCommand(commandList.info.note, notes.onNote);
+controller.registerCommand(commandList.info.all, notes.onAll);
+
+const officeHours = new OfficeHours(bot.client, sharedSettings, "data/office_hours_data.json");
+controller.registerCommand(commandList.officeHours.ask, officeHours.onAsk);
```
Querijn: Replace all of the command handlers with <class>.<handler>,bind(<class>)? Then all the functions can be class functions again, and they wont have to be arrow functions.

Querijn: eg in the case of officeHours.onAsk, it would become officeHours.onAsk.bind(officeHours)

PR #57, opened 2018-03-31
Add event handler

stelar7:
Should prevent the following message from being shown in console as we avoid calling .on()
>(node) warning: possible EventEmitter memory leak detected. 11 change listeners added. Use emitter.setMaxListeners() to increase limit.

Also avoids adding `bot` to every constructor..

### DIFF
--- a/settings/shared_settings.json
+++ b/settings/shared_settings.json
@@ -1,5 +1,6 @@
 {
     "server": "342988445498474498",
+    "name": "Botty McBotFace",
     "commands": {
         "default_prefix": "!",
         "adminRoles": [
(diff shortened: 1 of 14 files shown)


Querijn:
We could also call emitter.setMaxListeners...

stelar7:
:thinking:

Querijn:
Let's not make a giant overhaul because of a warning message..

stelar7:
I mean.. The overhaul is already done..

Querijn:
The main issue is, I have the responsibility of maintaining this after this PR. Maybe someone else takes over, and they have to. The added complexicity is just not worth the added cost to maintainability. I would much rather have a single call to client.setMaxListeners than this.

Not to demotivate you from doing this, I love the effort you put into this project, but let's discuss the things beforehand, or make an issue about them, this opens a discussion before I'd have to look through 300 changes. I really dislike closing this because a lot of effort went into it of course. I'm sorry.

PR #62, opened 2018-04-03
tag users in the question_list

stelar7:
Added since !ask and !ask_for stores different data in the list..

Optionally should !ask be changed to tag? as usernames might be different from the displayname..

### DIFF
--- a/src/OfficeHours.ts
+++ b/src/OfficeHours.ts
@@ -61,7 +61,7 @@ export default class OfficeHours {
 
     public onAsk(message: Discord.Message, isAdmin: boolean, command: string, args: string[]) {
         const question = args.join(" ");
-        this.storeQuestion(question, message, message.author.id, message.author.toString());
+        this.storeQuestion(question, message, message.author.toString(), message.author.username);
     }
 
     public onAskFor(message: Discord.Message, isAdmin: boolean, command: string, args: string[]) {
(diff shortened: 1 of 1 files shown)


Querijn:
Why not? If we cannot provide an answer, and we're going to ask Riot for it, they would probably like to be notified of the answer once they get it..

stelar7:
We should have !ask and !ask_for have the same interaction with the other commands imo. and as its now, one of them tags, the other doesnt.

Querijn:
I'm kind of confused where this PR is at right now, but basically, I need the following behaviour:
1. *Don't* mention users in !question_list, so that if it is used outside of a moderator channel, all the users don't get a notification
2. !ask and !ask_for I don't really care about, as long as they're consistent (both mention or neither mention)
3. The listing of users will be tagged during office hours.

Querijn:
At the least, this one needs to be fixed: 
afbeelding

stelar7:
As it is now, both !ask and !ask_for should add the mention to the datastore, if you dont want !question_list to tag users, we need to fetch users->username in that command.

src/OfficeHours.ts
```diff
@@ -61,15 +61,15 @@ export default class OfficeHours {
 
     public onAsk(message: Discord.Message, isAdmin: boolean, command: string, args: string[]) {
         const question = args.join(" ");
-        this.storeQuestion(question, message, message.author.id, message.author.toString());
+        this.storeQuestion(question, message, message.author.toString(), message.author.username);
     }
 
     public onAskFor(message: Discord.Message, isAdmin: boolean, command: string, args: string[]) {
         const asked = message.mentions.members.first();
         if (!asked) return;
```
Querijn: This makes sense in this boolean-esque situation, but is weird below (asked.id and asked.toString()). Can you change it to `asker`?

PR #64, opened 2018-04-11
Print a message in the channel where !question_list was used if the list is empty

stelar7:
Would the question list look better as an embed?


Querijn:
Yes I like it

PR #65, opened 2018-04-11
Send commands as an embed (improves formatting)

stelar7:


Querijn:
Might conflict with the choice to go for pagination, someone was working on that

PR #67, opened 2018-04-12
allow notes to have other text after them 

stelar7:
At the moment, doing `.note {some text here}` or `!note note {some text here}` makes botty ignore that command, this is not optimal for when you want to @ someone in the same mesasge, or provide context

### DIFF
--- a/src/Info.ts
+++ b/src/Info.ts
@@ -33,6 +33,7 @@ export default class Info {
 
     public onAll(message: Discord.Message, isAdmin: boolean, command: string, args: string[]) {
         let response: string | undefined;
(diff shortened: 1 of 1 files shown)


Querijn:
This has no other sideeffects with the other commands?

stelar7:
It will only reach onAll if you use .note or !note without a subcommand, so it should be fine

molenzwiebel:
If we want to be safe, change it to `args.length >= 1` (or `args.length > 0`)

stelar7:
It should never hit that point without a valid note name at args[0], so we dont need to care about the args at all.

Querijn:
Merged

PR #71, opened 2018-04-21
add common aliases for languages

### LINKED ISSUE #69: !lib aliasing
using !lib should accept common aliases, like .net, c#, js, ts

stelar7:
fixes #69

### DIFF
--- a/settings/shared_settings.json
+++ b/settings/shared_settings.json
@@ -52,7 +52,19 @@
         "languageList": "I have libraries for one of the following languages: {languages}\nChoose one you like and reply with `!libs <your language>`.",
         "githubErrorList": "I tried to get the languages from Github, but Github replied with status code ",
         "githubErrorLanguage": "I tried to get the language from Github, but Github replied with status code ",
-        "baseURL": "https://api.github.com/repos/WxWatch/riot-api-libraries/contents/libraries/"
+        "baseURL": "https://api.github.com/repos/WxWatch/riot-api-libraries/contents/libraries/",
+        "aliases": {
+            "javascript": [
+                "js",
+                "ts",
+                "typescript"
+            ],
+            "c-sharp": [
+                ".net",
+                "vb",
+                "c#"
+            ]
+        }
     },
     "autoReact": {
         "emoji": "408527155891273738"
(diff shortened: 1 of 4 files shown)


settings/shared_settings.json
```diff
@@ -52,7 +52,19 @@
         "languageList": "I have libraries for one of the following languages: {languages}\nChoose one you like and reply with `!libs <your language>`.",
         "githubErrorList": "I tried to get the languages from Github, but Github replied with status code ",
         "githubErrorLanguage": "I tried to get the language from Github, but Github replied with status code ",
-        "baseURL": "https://api.github.com/repos/WxWatch/riot-api-libraries/contents/libraries/"
+        "baseURL": "https://api.github.com/repos/WxWatch/riot-api-libraries/contents/libraries/",
+        "aliases": {
+            "javascript": [
+                "js",
+                "ts",
+                "typescript"
```
Querijn: 🤔

Querijn: Would not put this one in the js aliases

settings/shared_settings.json
```diff
@@ -52,7 +52,19 @@
...
+            ],
+            "c-sharp": [
+                ".net",
```
Querijn: Nor this one, they're just not the same

src/RiotAPILibraries.ts
```diff
@@ -129,9 +129,21 @@ export default class RiotAPILibraries {
...
 
-    private async getListForLanguage(message: Discord.Message, language: string) {
+    private async getListForLanguage(message: Discord.Message, language: string): Promise<void> {
         const response = await fetch(this.settings.riotApiLibraries.baseURL + language);
+
         if (response.status !== 200) {
+            const aliases = this.settings.riotApiLibraries.aliases;
+            const keys = Object.keys(aliases);
+
+            for (const index in keys) {
```
molenzwiebel: `for (const [, key] in Object.entries(aliases))`

stelar7: It doesnt like that 🤔 

`The left-hand side of a 'for...in' statement cannot be a destructuring pattern.`

stelar7: I cant get VSC to accept Object.entries in es6, so we would need to update to es2017? 🤔

Querijn: That is fine

PR #85, opened 2018-05-22
split schema update to its own class

### LINKED ISSUE #79: Extract Riot API Schema update to separate extension
Currently the Riot API Schema is requested only in the APIURlnterpreter. In the future it will also be used in the endpoint feature. 

Requesting, updating and caching the API Schema should be in its own class. This way multiple extensions can rely on an up-to-date schema without reimplementing the Schema HTTP Request multiple times.
The new extension then informs all interested classe about schema updates (Observer Pattern).

stelar7:
fixes #79


Querijn:
Merge conflicts

Issue #76, opened 2018-05-21
Simplify linking to API endpoints


TiFu:
Currently linking to API endpoints like `match` or `summoner` requires going to the dev portal and copy pasting the link to it. 
I'd like to add a new command `!endpoint summoner|match|....` which automatically posts the link to the endpoint.

One could also think about trying to automatically detect sentences like `use the summoner endpoint` and then automatically send a link to it. This feature would need a cooldown though to avoid spamming the chat.

Which version should be implemented?

I'll start working on the basic structure today and then adapt it depending on the chosen version.

Querijn:
Rate limiting should be its own feature inside the command handler possibly. Endpoint command seems like a good idea! Potentially built on the riot api schema? 

https://github.com/Querijn/BottyMcBotface/blob/API-URL-Fetch/src/ApiUrlInterpreter.ts#L384

TiFu:
Sounds good. Thanks for the input!

PR #88, opened 2018-06-02
Implemented endpoint command

TiFu:
See #76. 
This PR introduces a new `!endpoint [endpoint name]` command. The bot then answers with a direct link to the endpoint e.g. 

Command:
```
!endpoint status
```
Response:
```
@TiFu, https://developer.riotgames.com/api-methods/#lol-status-v3
```


settings/command_list.json
```diff
@@ -136,5 +136,10 @@
             "apistatus"
         ],
         "description": "Prints the status of the Riot Games API"
+    },
+    "endpoint": {
+        "aliases": [
+            "endpoint"
```
Querijn: aliases for itself are not necessary.

Querijn: iirc

src/Endpoint.ts
```diff
@@ -0,0 +1,186 @@
...
+            message.reply("Could not find requested endpoint!");
+        }
+    }
+
+    /**
+     * Levenshtein Distance with some simple modifications for endpoint & input
+     * @param endpoint
+     * @param input
+     */
+    private calculateDistance(endpoint: string, input: string): number {
```
Querijn: We have this function somewhere in the codebase already, can we combine if so?

src/Endpoint.ts
```diff
@@ -0,0 +1,186 @@
...
+        endpoint = endpoint.replace(new RegExp("-", "g"), " ");
+        input = input.replace(new RegExp("-", "g"), " ");
+
+        const d = new Array(endpoint.length + 1);
```
Querijn: Is this an array with an expected length?

src/Endpoint.ts
```diff
@@ -0,0 +1,186 @@
...
+            const paths = schema.paths;
+
+            const endpointSet = new Set<EndpointName>();
+
+            for (const path in paths) {
+                const endpointName = paths[path]["x-endpoint"]; // match-v3
+                endpointSet.add(endpointName);
+            }
+
+            this.endpoints.lastUpdate = new Date();
```
Querijn: Unused?

src/Endpoint.ts
```diff
@@ -0,0 +1,186 @@
...
+
+            if (response.status !== 200) {
+                console.error("HTTP Error trying to get schema: " + response.status);
+                return;
+            }
+
+            const schema = await response.json();
+            const paths = schema.paths;
+
+            const endpointSet = new Set<EndpointName>();
```
Querijn: Could you explain Set to me?

src/Endpoint.ts
```diff
@@ -0,0 +1,186 @@
...
+        this.updateEndpoints();
+    }
+
+    public onEndpoint(message: Discord.Message, isAdmin: boolean, command: string, args: string[]) {
+        if (this.endpoints.endpoints === undefined) {
+            message.reply("Sorry, endpoints are not yet initialized!");
+            return;
+        }
+        const argsString = args.reduce((prev, current) => prev + current, "");
+        let minDist = Infinity;
```
Querijn: And here I always use 9e9.. lol, til

Querijn:
#85 is potentially related

settings/command_list.json
TiFu: The bot does not respond to `!endpoint status` if I remove it from aliases (just tested again - that's also the reason why I added it in the first place)

src/Endpoint.ts
TiFu: I don't want any duplicate endpoints. (i.e. if an endpoint has two routes (https://developer.riotgames.com/api-methods/#champion-v3) I get `champion` twice in the loop below.
I have to copy to an array because fileBackedObject does not properly serialize sets.

src/Endpoint.ts
TiFu: Good catch.

src/Endpoint.ts
TiFu: It is because levenshtein distance needs an array of `s1.length` + 1 times `s2.length + 1` for the input strings `s1` and `s2`

TiFu:
Fixed the issues you mentioned. 

If I understand the @stelar7 code changes correctly (only looked through them very quickly), the new class for schema updates is still specific to the APIUrlInterpreter .
I'd rather see the schema update be its own class (Observable) which only handles the updates - the specific stuff which a module does with the schema should stay in that module (e.g. the way APIUrlInterpreter parses the schema).

```
class SchemaUpdater {
     private schemaObservers: SchemaObserver[]

     // other methods e.g. schema update admin command
     addSchemaObserver(schemaObserver) {
              this.schemaObservers.push(schemaObserver);
      }

    updateSchema() {
             if (schemaChanged)
                     this.notifyObservesr(newSchema);
     }


      notifyObservers(newSchema) {
              for (let schemaObserver in this.schemaObservers) {
                    schemaObserver.onSchemaUpdated(newSchema)
              }
       }
}

interface SchemaObserver {
         onSchemaUpdated(schema: any): void
}
```

src/Endpoint.ts
```diff
@@ -182,5 +142,14 @@ export default class Endpoint {
...
             throw e;
         }
+
+        if (this.timeOut) {
+            clearTimeout(this.timeOut);
+            this.timeOut = null;
+        }
+
+        // somehow TS 2.8.3 doesn't like this.timeOut = setTimeout(this.updateEndpoints.bind(this), this.timeOutDuration);
+        // (typeof this.timeout == Timer | null vs typeof setTimeout() == number...)
```
stelar7: I think I had the same issue, but it was solved by adding
`import { setTimeout } from "timers";`

TiFu: Thank you fixed with 527b5249fe577fab791246ce5a42a9fa8a212b49

PR #93, opened 2018-06-15
Change error messages to be more informative

stelar7:


src/ForumReader.ts
```diff
@@ -58,15 +58,16 @@ export default class ForumReader {
         bot.on("ready", () => {
             const guild = bot.guilds.get(this.sharedSettings.server);
             if (!guild) {
-                console.error(`ForumReader: Incorrect settings for guild ID ${this.sharedSettings.server}`);
+                console.error(`ForumReader: Unable to find server: ${this.sharedSettings.server}`);
```
Querijn: Would still point out that they're IDs, this could be interpreted as server names

PR #96, opened 2018-07-11
Create !version command

stelar7:
Not sure how the bot is started, but if its done with `git pull && npm install && npm run` (or something similar), then this should work fine.


src/app.ts
```diff
@@ -48,9 +48,11 @@ controller.registerCommand(commandList.controller.help, controller.onHelp.bind(c
 
 // Botty commands
 controller.registerCommand(commandList.botty.restart, bot.onRestart.bind(bot));
+controller.registerCommand(commandList.botty.version, bot.onVersion.bind(bot));
 
 // API schema commands
 controller.registerCommand(commandList.apiSchema.updateSchema, apiSchema.onUpdateSchemaRequest.bind(apiSchema));
+controller.registerCommand(commandList.apiUrlInterpreter.updateSchema, apiUrlInterpreter.onUpdateSchemaRequest.bind(apiUrlInterpreter));
```
stelar7: not sure what line is the correct one atm, it appeared as a merge issue so i just added both :LUL:

src/Botty.ts
```diff
@@ -18,11 +18,14 @@ export default class Botty {
...
     private personalSettings: PersonalSettings;
     private sharedSettings: SharedSettings;
+    private version: string;
 
     constructor(sharedSettings: SharedSettings) {
         this.personalSettings = sharedSettings.botty;
         this.sharedSettings = sharedSettings;
         console.log("Successfully loaded bot settings.");
+        
+        require("fs").readFile("version", "utf-8", (err,data) => this.version = data);
```
molenzwiebel:
```
this.version = fs.readFileSync("version", "utf8");
```
Probably cleaner to import fs instead of doing a require. Typescript types an `import` properly, while it types `require` as any.

settings/command_list.json
```diff
@@ -1,5 +1,11 @@
 {
     "botty": {
+        "version": {
+            "aliases": [
+                "version"
+            ],
+            "description":"Prints the git version the bot was built from"
```
molenzwiebel: SPACES AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA

saveVersion.js
```diff
@@ -0,0 +1,3 @@
+var spawn = require("child_process").spawn;
+var git = spawn("git" ["rev-parse", "HEAD"]);
```
molenzwiebel: `const`

molenzwiebel: also, what the fuck? did you never test this lmao. There's a `,` missing between `"git"` and the arguments

src/Botty.ts
```diff
@@ -97,6 +100,10 @@ export default class Botty {
         });
         console.log("Initialised listeners.");
     }
+    
+     public onVersion(message: Discord.Message, isAdmin: boolean, command: string, args: string[]) {
+        message.channel.send(this.version);
```
molenzwiebel: Maybe something like `"Current Version: " + version` instead of just dumping the version in chat?

saveVersion.js
Querijn: It's also javascript (js file)

Why was this not incorporated into the class?

Querijn:
There's a versioning with the tags which is what I would like to use for the fetching, loading, compiling, etc
Then we could check what version we're on by checking that. This needs more thought put into it.

PR #99, opened 2018-07-20
Improve server stability

s-coimbra21:
Improves server stability by adding a couple maintenance tasks that prevent the server from falling into disarray.


Querijn:
Merged

PR #104, opened 2018-08-04
Improve API call URL error detection

Derpthemeus:
I overhauled the API call identification and error detection extension. It would be greatly appreciated if somebody wants to do a code review, and also test any edge cases you can think of.

Enhancements:
- Works correctly when query parameters are specified
- Identifies missing required query parameters
- Identifies query parameters that are the incorrect type
- Identifies query parameters that do not exist for the used endpoint
- Identifies calls that use HTTP instead of HTTPS
- Can identify multiple API calls in one message
- Doesn't make API calls that are expected to fail
- Works with tournament-stub-v3

### DIFF
--- a/src/ApiSchema.ts
+++ b/src/ApiSchema.ts
@@ -1,46 +1,54 @@
 import Discord = require("discord.js");
 import fetch from "node-fetch";
+import XRegExp = require("xregexp");
 
 import levenshteinDistance from "./LevenshteinDistance";
 import { SharedSettings } from "./SharedSettings";
 
 import { clearTimeout, setTimeout } from "timers";
 
-class PathRegexCollection {
-    public invalid: string;
-    public valid: string;
-
-    constructor(validString: string, invalidString: string) {
-        this.valid = validString;
-        this.invalid = invalidString;
-    }
-}
-
-class SchemaRegexCollection extends PathRegexCollection {
-    public name: string;
-    public schema: any;
-
-    constructor(name: string, schema: any, validString: string, invalidString: string) {
-        super(validString, invalidString);
-
-        this.name = name;
-        this.schema = schema;
-    }
-}
-
 export class Path {
-    public methodType: "GET" | "POST";
-    public regex: PathRegexCollection;
-    public parameterInfo: SchemaRegexCollection[];
-    public method: string;
     public name: string;
+    public methodType: "GET" | "POST";
+    public regex: RegExp;
+    public pathParameters: Map<string, Parameter>;
+    public queryParameters: Map<string, Parameter>;
+}
+
+export class Parameter {
+    public name: string;
+    public required: boolean;
+    public type: ParameterType;
+}
+
+export class ParameterType {
+    /** A human readable description (e.g. "a positive integer") */
+    public description: string;
+
+    constructor(description: string, isValidValue: (value: string) => boolean){
+        this.description = description;
+        this.isValidValue = isValidValue;
+    }
+    /** A function that returns a boolean indicating if the specified value is a valid value for this type of parameter */
+    public isValidValue(value: string | string[]): boolean { return false; }
 }
 
 export class APISchema {
 
+    private static PARAMETER_TYPES = {
+        ANY: new ParameterType("anything", (value) => true),
+        STRING: new ParameterType("a string", (value) => !Array.isArray(value)),
+        INTEGER: new ParameterType("an integer", (value) => Number.isInteger(+value)),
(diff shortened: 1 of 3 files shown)


src/ApiSchema.ts
```diff
@@ -1,46 +1,54 @@
...
 }
 
 export class APISchema {
 
+    private static PARAMETER_TYPES = {
+        ANY: new ParameterType("anything", (value) => true),
+        STRING: new ParameterType("a string", (value) => !Array.isArray(value)),
+        INTEGER: new ParameterType("an integer", (value) => Number.isInteger(+value)),
+        BOOLEAN: new ParameterType("a boolean", (value) => value === "true" || value === "false"),
+        SET: new ParameterType("a set specified like `paramName=value1&paramName=value2`", (value) => {
```
molenzwiebel: Technically you can just do value without thr braces. More a stylistic choice though. I think we mostly have it without braces in this codebase?

src/ApiSchema.ts
```diff
@@ -95,23 +103,34 @@ export class APISchema {
...
 
             for (const pathName in schema.paths) {
                 const pathSchema = schema.paths[pathName];
                 const methodSchema = pathSchema.get ? pathSchema.get : pathSchema.post;
 
                 if (!methodSchema) continue; // Only handle GET/POST
-                if (methodSchema.operationId.startsWith("tournament")) continue;
+                if (methodSchema.operationId.startsWith("tournament-v3")) continue;
 
                 const path = new Path();
```
molenzwiebel: imo a fully fledged constructor instead of this building would be better suited here. You can just initialize the parameter maps to empty maps.

src/ApiSchema.ts
```diff
@@ -95,23 +103,34 @@ export class APISchema {
...
-                this.constructRegex(path, baseUrlRegex, invalidBaseUrlRegex, pathName, methodSchema);
-                console.assert(path.regex, "Path Regex should be set");
+                path.regex = this.constructRegex(pathName, methodSchema);
+
+                path.queryParameters = new Map();
+                path.pathParameters = new Map();
+                if (methodSchema.parameters) {
+                    for (const parameter of methodSchema.parameters) {
+                        if (parameter.in === "path") {
+                            this.addParameter(path.pathParameters, parameter);
```
molenzwiebel: imo Path.addParameter(..) makes more sense than this.addParameter(..), especially from an oop standpoint

src/ApiSchema.ts
```diff
@@ -129,78 +148,42 @@ export class APISchema {
...
+        regex = regex.replace(/\{(.*?)\}/, (match, p2) => `(?<${p2}>[^\\/?\\s]*)`);
+
+        return XRegExp(regex);
     }
 
-    private constructRegex(path: Path, validBase: string, invalidBase: string, pathName: string, methodSchema: any) {
-
-        path.name = pathName;
-        path.parameterInfo = [];
+    private addParameter(map: Map<string, Parameter>, parameterSchema: any) {
```
molenzwiebel: Like mentioned, consider converting this to a Path member function.

src/ApiSchema.ts
```diff
@@ -129,78 +148,42 @@ export class APISchema {
...
+        const parameter = new Parameter();
+        parameter.name = parameterSchema.name;
+        parameter.required = parameterSchema.required;
+        parameter.type = this.getParameterType(parameterSchema.schema.type);
```
molenzwiebel: Same goes with the class filling instead of a proper constructor here. Or you can just make Parameter an interface instead of a class and use plain objects

src/ApiUrlInterpreter.ts
```diff
@@ -65,89 +75,115 @@ export default class ApiUrlInterpreter {
...
+    private async testRiotApiUrl(url: string, message: Discord.Message) {
+        // The type must be `any` because the regex contains named groups
+        const urlMatch: any = XRegExp.exec(url, ApiUrlInterpreter.API_CALL_REGEX);
 
-            // Check if path was valid
-            const validMatch = new RegExp(path.regex.valid, "g").exec(content);
-            if (validMatch && validMatch.length > 0) {
+        /** Indicates if there is a problem with this URL that guarantees the call will fail. */
+        let fatalError: boolean = false;
+        const mistakes = [];
```
molenzwiebel: You shouldn’t need to specify : boolean here

Derpthemeus:
Thanks for the feedback! PR has been updated and merged.

PR #112, opened 2018-08-19
only react if the last message was over an hour ago

stelar7:


stelar7:
Turns out, D.JS doesnt allow you to fetch the last message a user posted in a cleaner way than fetching the entire chatlog and parsing that. That would be out of scope for a feature like this imo. and thusly fuckit.

PR #113, opened 2018-09-03
Various fixes to the esports api and the private settings file

stelar7:
Dont print when there are no games
Check for updates to the api before every call (with cache)

add missing fields to the private settings to avoid errors because of missing fields

### DIFF
--- a/private/shared_settings.json
+++ b/private/shared_settings.json
@@ -3,16 +3,28 @@
     "botty": {
         "discord": {
             "key": "",
-            "owner": 95265940081475584
+            "owner": 132268173432061952
+        },
+        "honeypot": {
+            "token": "",
+            "owner": 132268173432061952
         },
         "forum": {
             "username": "<lowercase summoner name> (<region>)",
             "password": ""
         },
+        "riotApi": {
+            "key": ""
+        },
+        "webServer": {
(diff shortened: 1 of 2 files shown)


src/ESports.ts
```diff
@@ -124,9 +127,9 @@ export default class ESportsAPI {
 
     private sendPrintout(channel: Discord.TextChannel, data: Map<string, ESportsLeagueSchedule[]> | undefined, date: string) {
 
-        date = date.split(" ").join("/");
-        if (!data) {
-            channel.send(`No games played on ${date}.`);
+        date = (date || momentjs().format("YYYY M D")).split(" ").join("/");
+        if (!data || data.size === 0) {
```
molenzwiebel: `data.length`, but `data` will already evaluate to false if the length is `0`, so this is duplicate and won't be executed.

stelar7: data is a Map, and an empty map does not evaluate to false. Thus, this will not be changed.

private/shared_settings.json
```diff
@@ -3,16 +3,28 @@
     "botty": {
         "discord": {
             "key": "",
-            "owner": 95265940081475584
+            "owner": 132268173432061952
```
Querijn: @stelar7

PR #125, opened 2018-10-18
move pickem to a separate class and add brackets

stelar7:
user lookup: https://i.imgur.com/rcygnTh.png
current : https://i.imgur.com/RDLbRee.png
leaderboard: https://i.imgur.com/pO1VHqS.png


Querijn:
Can you add Pickems to the readme document?
Also Esports, since it's apparently missing too

PR #127, opened 2018-10-21
Add remaining settings to the config

stelar7:
Since the private file overwrites the public file, the missing params from the private file removes whatever you have in the public config.

### DIFF
--- a/private/shared_settings.json
+++ b/private/shared_settings.json
@@ -13,6 +13,14 @@
             "username": "Querijn",
             "password": ""
         },
-        "isProduction": false
+        "riotApi": {
+            "key": ""
+        },
+        "webServer": {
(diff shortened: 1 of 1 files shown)


Querijn:
What's the use of this?

stelar7:
Not having this makes the app unable to start because of missing key errors.

Querijn:
But they're in the public file, right?

stelar7:
Yes, but the private overwrites the public, so since they are missing from the private, they are also missing in the loaded config.

It needs to be changed here (an easy fix)
or here 
https://github.com/Querijn/BottyMcBotface/blob/master/src/FileBackedObject.ts#L10-L16

Querijn:
I would prefer that.

PR #134, opened 2018-10-27
Use evenin to detect greetings

molenzwiebel:
Got tired of having a hardcoded list of greetings for only a certain amount of languages. Now uses [evenin](https://github.com/molenzwiebel/evenin), which supports slightly more phrases.

If we're interested, might also add a command to look up what a greeting means.

Also we should probably have a blacklist for some greetings since I assume that some are going to get very annoying very fast.

<3

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "bufferutil": "^3.0.0",
     "cheerio": "^1.0.0-rc.2",
     "discord.js": "11.3.2",
+    "evenin": "^1.0.0",
     "feed-read": "^0.0.1",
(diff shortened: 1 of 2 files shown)


Querijn:
The type definitions are missing from the package, at least when I install it

Querijn:
hasGreeting doesn't work in cases where the words aren't split up by spaces, needs punctuation as well.

Querijn:
Alright, I've tested it enough, it works but evenin needs a couple of changes. I'm going to make a PR for the last issue.

Azoy:
smh

PR #144, opened 2018-11-03
fix timer type and add evenin definitions

stelar7:


Querijn:
evenin is no longer being used
the timer type belongs to the other pr

PR #145, opened 2018-11-03
fix notes not showing up in the list

### LINKED ISSUE #109: Newly added notes won't be in !note list
This is because the old emojis are listed like `<emoji:id>` while newly-added ones are stored as `id`

stelar7:
fixes #109

### DIFF
--- a/src/Info.ts
+++ b/src/Info.ts
@@ -231,7 +231,7 @@ export default class Info {
             command,
             counter: 0,
             message,
-            categoryId: category.id,
+            categoryId: category.id || category.identifier,
         };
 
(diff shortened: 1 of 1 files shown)


Querijn:
nice, was looking forward to this one

stelar7:
it works for me 🤷‍♂️ the id seems to always be null

Querijn:
Maybe because it needs to be fetched.. In that case, it can be removed altogether, I assume?

stelar7:
Dont know, havent tested that

Querijn:
If you could confirm, that would be amazing, and I could upload a new version of Botty today

PR #151, opened 2018-11-25
add TTT as a game

stelar7:
this is prolly the worst code ive ever written 👍


src/games/TicTacToe.ts
```diff
@@ -0,0 +1,248 @@
...
+                return;
+            }
+        }
+        if (game.turnPlayer === TicTacToeType.O) {
+            if (message.author.id === game.Xplayer) {
+                return;
+            }
+        }
+
+        const regex = /\d/;
```
molenzwiebel: `/^\d$/`, otherwise a digit located anywhere in the message is accepted, then conversion to int fails

src/games/TicTacToe.ts
```diff
@@ -0,0 +1,248 @@
...
+            }
+        }
+    }
+
+    private isValidMove(board: { [key: number]: TicTacToeType }, move: number) {
+        return board[move] === undefined;
+    }
+
+    private incrementGameState(game: TicTacToeGame, move: number) {
+        if (move) {
```
molenzwiebel: You call this with -1 in the invite handler. -1 is thruthy so you will set the -1th element of the board, which shouldn't be the intention. move > 0 should be good.

src/games/TicTacToe.ts
```diff
@@ -0,0 +1,248 @@
...
+        if (isBottyOpponent) {
+            message.reply("Playing against Botty does not work at the moment.. :(");
+            return;
+        }
+
+        const senderName = message.author.username;
+        const opponentName = opponent.username;
+        const key = message.author.id.localeCompare(opponent.id) ? message.author.id + "_" + opponent.id : opponent.id + "_" + message.author.id;
+        const highP = message.author.id.localeCompare(opponent.id) ? senderName : opponentName;
+        const lowP = opponent.id.localeCompare(message.author.id) ? opponentName : senderName;
```
molenzwiebel: These lines do not do what you think they do. localeCompare returns either a negative number, zero or a positive number. The function returns zero only if the values are the same, and since both positive and negative numbers are truthy, the first branch will always be taken (unless the player plays against themselves).

src/games/TicTacToe.ts
```diff
@@ -0,0 +1,248 @@
...
+        if (move) {
+            game.board[move] = game.turnPlayer;
+        }
+
+        game.turnPlayer = TicTacToeType.opposite(game.turnPlayer)!;
+        this.printGameBoard(game);
+
+        const winner = this.getGameWinner(game);
+        if (winner) {
+            const high = game.Oplayer.localeCompare(game.Xplayer);
```
molenzwiebel: Again, localeCompare

src/games/TicTacToe.ts
```diff
@@ -0,0 +1,248 @@
...
+            if (winner === game.Oplayer) {
+                if (high) {
+                    this.scores[game.scoreKey].highWins = this.scores[game.scoreKey].highWins + 1;
```
molenzwiebel: Just use ++ or += 1

src/games/TicTacToe.ts
```diff
@@ -0,0 +1,248 @@
...
+
+            const highP = high ? oplayer.user.username : xplayer.user.username;
+            const lowP = high ? xplayer.user.username : oplayer.user.username;
+            const highScore = this.scores[game.scoreKey].highWins;
+            const lowScore = this.scores[game.scoreKey].lowWins;
+            const drawScore = this.scores[game.scoreKey].draws;
+
+            game.channel.send("The game has ended!\nScores are now: " + highP + "(" + highScore + ") - " + drawScore + " - " + lowP + "(" + lowScore + ")");
+            game.channel.overwritePermissions(oplayer, { SEND_MESSAGES: false });
+            game.channel.overwritePermissions(xplayer, { SEND_MESSAGES: false });
```
molenzwiebel: Channel is never deleted, will clutter bigtime

src/games/TicTacToe.ts
```diff
@@ -0,0 +1,248 @@
...
+            (game.board[7] === player && game.board[8] === player && game.board[9] === player) ||
+            (game.board[1] === player && game.board[4] === player && game.board[7] === player) ||
+            (game.board[2] === player && game.board[5] === player && game.board[8] === player) ||
+            (game.board[3] === player && game.board[6] === player && game.board[9] === player) ||
+            (game.board[1] === player && game.board[5] === player && game.board[9] === player) ||
+            (game.board[3] === player && game.board[5] === player && game.board[7] === player)) {
+            return game.Oplayer;
+        }
+
+        for (let i = 1; i < 10; i++) {
```
molenzwiebel: if (game.board.some(x=>!x)) return;

return is the same as return undefined

stelar7: board is a "map" type, not an array, so .some() is not an option

src/games/TicTacToe.ts
```diff
@@ -0,0 +1,400 @@
...
+            this.handleEndOfGame(game, game.Oplayer.id);
+        }
+
+        if (sender === game.Oplayer.id) {
+            this.handleEndOfGame(game, game.Xplayer.id);
+        }
+    }
+
+    private handleDraw(game: TicTacToeGame, sender: string) {
+        if (sender === game.Xplayer.id) {
```
molenzwiebel: Ever considered doing a for loop with 2 entries instead of duplicating code.

This goes for more than just this block

src/games/TicTacToe.ts
```diff
@@ -0,0 +1,400 @@
...
+            if (game.board[i] === undefined) {
+                return undefined;
+            }
+        }
+
+        return "-1";
+    }
+
+    private testPlayerWin(game: TicTacToeGame, player: TicTacToeType) {
+        return (game.board[1] === player && game.board[2] === player && game.board[3] === player) ||
```
molenzwiebel:
```js
const c = (...args) => args.filter(x => game.board[x] === player).length === args.length;
return c(1, 2, 3) || c(4, 5, 6) || etc
```

src/games/TicTacToe.ts
```diff
@@ -0,0 +1,370 @@
...
+            if (game.board[i] === undefined) {
+                return undefined;
+            }
+        }
+
+        return "-1";
+    }
+
+    private testPlayerWin(game: TicTacToeGame, player: TicTacToeType) {
+        const isAllPlayer = (...args: number[]) => args.filter(x => game.board[x] === player).length === args.length;
```
molenzwiebel: Can probably just use `.every` here instead of `filter(x) === length`, but thats my bad.

PR #187, opened 2019-03-26
Add game data searching functionality to botty

stelar7:


src/GameData.ts
```diff
@@ -0,0 +1,292 @@
...
+                levenshteinDistance(search, c.name),
+                levenshteinDistance(search, c.key),
+            ]) < 3);
+
+        if (result.length === 0) {
+            return "Unable to find any data for that search term.";
+        }
+
+        if (result.length > 1) {
+            let response = "Too many results returned for that search, please try another;\n";
```
Querijn: ; in this string?

src/GameData.ts
```diff
@@ -0,0 +1,292 @@
...
+            result.forEach(r => response += `\`${r}\`, `);
+            response = response.slice(0, -2);
+            return response;
+        }
+
+        const returnedValue = {
+            id: result[0].id,
+            name: result[0].name,
+            endOfGameStatDescs: result[0].endOfGameStatDescs,
+        } as any;
```
Querijn: is this necessary?

src/GameData.ts
stelar7: Ment as a list separator, but apparently you are supposed to use : if the list has more than 3 elements, so its changed now

src/GameData.ts
stelar7: Not at this location, its a copy-paste "error"

src/GameData.ts
```diff
@@ -0,0 +1,265 @@
...
+            });
+
+            i.to.forEach((f: number) => {
+                const otherItem = returnData.filter(r => r.id === f)[0];
+                if (otherItem !== undefined) {
+                    item.to.push(otherItem.name);
+                }
+            });
+        });
+
```
molenzwiebel:
```js
for (const item of items) {
    const complex = returnData.find(x => x.id === item.id)!;
   
    complex.from = item.from.map(x => returnData.find(y => y.id === x)!).filter(x => x).map(x => x.name);
    complex.to = item.to.map(x => returnData.find(y => y.id === x)!).filter(x => x).map(x => x.name);
}
```

src/GameData.ts
```diff
@@ -0,0 +1,265 @@
...
+        return returnData;
+    }
+
+    public onLookup(message: Discord.Message, isAdmin: boolean, command: string, args: string[]) {
+        if (message.cleanContent.length === 0) {
+            let response = `I have info on the following categories; \`item\`,\`perk\`,\`champion\``;
+            response += `type !${command} {search_type} {search_term} to use it`;
+            message.channel.send(response);
+            return;
```
molenzwiebel:
```js
const response = `Usage: !${command} [type] [term]. Supported types are \`item\`, \`perk\` and \`champion\`.`;
```

src/GameData.ts
```diff
@@ -0,0 +1,265 @@
...
+        }
+
+        if (!["item", "perk", "champion"].some(i => i === args[0])) {
```
molenzwiebel:
```suggestion
        if (!["item", "perk", "champion"].includes(args[0])) {
```

src/GameData.ts
```diff
@@ -0,0 +1,265 @@
...
+            message.channel.send(`I'm sorry. I'm unable to parse the category \`${args[0]}\` at this moment. If you want it added, contact a guru`);
```
molenzwiebel: "If you think it should be added, please contact a guru." (better grammar, actual period at the end of the sentence)

src/GameData.ts
```diff
@@ -0,0 +1,265 @@
...
+        }
+    }
+
+    public findItem(search: string): string {
+        if (!search) {
+            return `There are currently ${this.itemData.length} items in my lookup data!`;
+        }
+
+        const result = this.itemData.filter(c => Math.min(
+            ...[
```
molenzwiebel: Unnecessary spread here. You can simply do `Math.min(a, b, c)`.

src/GameData.ts
```diff
@@ -0,0 +1,265 @@
...
+                levenshteinDistance(search, c.id.toString()),
+                levenshteinDistance(search, c.name),
+                levenshteinDistance(search, c.categories.toString()),
+            ]) < 1);
```
molenzwiebel: 1 may be a bit strict. Are there a lot of conflicting matches if you increase it?

src/GameData.ts
```diff
@@ -0,0 +1,265 @@
...
+
+        if (result.length === 0) {
+            return "Unable to find any data for that search term.";
+        }
+
+        if (result.length > 3) {
+            let response = "Too many results returned for that search, please try one of the options below:\n";
+            result.forEach(r => response += `\`${r.name}\`, `);
+            response = response.slice(0, -2);
+            return response;
```
molenzwiebel:
```js
response += result.map(x => "`" + r.name + "`").join(", ");
```

src/GameData.ts
```diff
@@ -0,0 +1,265 @@
...
+        const returnedValue = {
+            id: result[0].id,
+            name: result[0].name,
+            combineCost: result[0].price,
+            cost: result[0].priceTotal,
+            from: result[0].from,
+            to: result[0].to,
+        };
+
+        return "```" + JSON.stringify(returnedValue, null, 4) + "```";
```
molenzwiebel: Why isn't this just `JSON.stringify(result[0], null, 4)`? Why are you creating a shallow clone?

src/GameData.ts
```diff
@@ -0,0 +1,265 @@
...
+        if (!search) {
+            return `There are currently ${this.perkData.length} perks in my lookup data!`;
+        }
+
+        const result = this.perkData.filter(c => Math.min(
+            ...[
+                levenshteinDistance(search, c.id.toString()),
+                levenshteinDistance(search, c.name),
+                levenshteinDistance(search, c.shortDesc),
+            ]) < 1);
```
molenzwiebel: No spread, 1 may be a bit low. See earlier comments.

src/GameData.ts
```diff
@@ -0,0 +1,265 @@
...
+            message.channel.send(`I'm sorry. I'm unable to parse the category \`${args[0]}\` at this moment. If you want it added, contact a guru`);
+            return;
+        }
+
+        if (args[0] === "item") {
+            message.channel.send(this.findItem(args.slice(1).join(" ")));
+            return;
+        }
+
+        if (args[0] === "perk") {
```
molenzwiebel: Rune(s) as an alias for perk?

src/GameData.ts
```diff
@@ -0,0 +1,265 @@
...
+                levenshteinDistance(search, c.id.toString()),
+                levenshteinDistance(search, c.name),
+                levenshteinDistance(search, c.categories.toString()),
+            ]) < 1);
+
+        if (result.length === 0) {
+            return "Unable to find any data for that search term.";
+        }
+
+        if (result.length > 3) {
```
molenzwiebel: If there's 1-3 items, you only return the first one and don't show an error. Either this should be 1 or you should return the other matches too.

src/GameData.ts
```diff
@@ -0,0 +1,265 @@
...
+
+        if (result.length === 0) {
+            return "Unable to find any data for that search term.";
+        }
+
+        if (result.length > 1) {
+            let response = "Too many results returned for that search, please try one of the options below:\n";
+            result.forEach(r => response += `\`${r}\`, `);
+            response = response.slice(0, -2);
+            return response;
```
molenzwiebel: map + join instead of manually building the string and then splicing. See earlier comments.

src/GameData.ts
```diff
@@ -0,0 +1,265 @@
...
+        }
+
+        const returnedValue = {
+            id: result[0].id,
+            name: result[0].name,
+            endOfGameStatDescs: result[0].endOfGameStatDescs,
+        };
+
+        return "```" + JSON.stringify(returnedValue, null, 4) + "```";
```
molenzwiebel: Again, why a shallow clone?

src/GameData.ts
```diff
@@ -0,0 +1,265 @@
...
+        if (!search) {
+            return `There are currently ${this.champData.length} champions in my lookup data!`;
+        }
+
+        const result = this.champData.filter(c => Math.min(
+            ...[
+                levenshteinDistance(search, c.id.toString()),
+                levenshteinDistance(search, c.name),
+                levenshteinDistance(search, c.key),
+            ]) < 3);
```
molenzwiebel: ditto

src/GameData.ts
```diff
@@ -0,0 +1,265 @@
...
+
+        if (result.length === 0) {
+            return "Unable to find any data for that search term.";
+        }
+
+        if (result.length > 1) {
+            let response = "Too many results returned for that search, please try one of the options below:\n";
+            result.forEach(r => response += `\`${r}\`, `);
+            response = response.slice(0, -2);
+            return response;
```
molenzwiebel: ditto

src/GameData.ts
```diff
@@ -0,0 +1,265 @@
...
+        }
+
+        // maybe return the skin names here?
+        const returnedValue = { ...result[0] } as any;
+        returnedValue.skins = returnedValue.skins.length;
+
+        return "```" + JSON.stringify(returnedValue, null, 4) + "```";
```
molenzwiebel:
```suggestion
        return "```" + JSON.stringify({
            ...result[0],
            skins: result[0].skins.length
        }, null, 4) + "```";
```

src/GameData.ts
```diff
@@ -0,0 +1,265 @@
...
+                    .filter(k => k.startsWith(id))
+                    .filter(k => (k.length - id.length) === 3)
+                    .map(k => +k);
+
+                skinKeys.forEach(key => {
+                    const data = skins[key];
+                    returnData.filter(r => r.id === +id)[0].skins.push({ ...data });
+                });
+            });
+
```
molenzwiebel:
```js
for (const champion of returnData) {
	champion.skins = skins.filter(x => Math.floor(x.id / 1000) === champion.id);
}
```

src/GameData.ts
```diff
@@ -0,0 +1,292 @@
...
+    private itemData: ItemData[];
+
+    private bot: Discord.Client;
+
+    public constructor(bot: Discord.Client) {
+        this.bot = bot;
+
+        bot.on("ready", async () => {
+            this.reloadData();
+            setInterval(this.reloadData, 86400 * 1000);
```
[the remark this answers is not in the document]
idunnololz: Make a const for this instead. Write the interval as a comment for the const.

src/GameData.ts
idunnololz: I like perk better. Ignore this.

src/GameData.ts
```diff
@@ -0,0 +1,240 @@
...
+
+        return "```" + JSON.stringify(result[0], null, 4) + "```";
+    }
+
+    public async findChampion(search: string) {
+        if (!search || search === undefined || search == null || search.length === 0) {
+            return "There are currently ${champData.length} champions on the PBE server!";
+        }
+
+        const result = this.champData.filter(c => Math.max(levenshteinDistance(search, c.id.toString()), levenshteinDistance(search, c.name), levenshteinDistance(search, c.key)) < 3);
```
idunnololz: Line 2 longggggggggggg

idunnololz:
```
const result = this.champData
    .filter(c => 
        3 > Math.max(
            levenshteinDistance(search, c.id.toString()), 
            levenshteinDistance(search, c.name), 
            levenshteinDistance(search, c.key)
        )
   );
```

src/GameData.ts
idunnololz: Hey I know one liners look cool but have you tried `making your code readable`?

src/GameData.ts
stelar7: Outdated

src/GameData.ts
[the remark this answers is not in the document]
idunnololz: E.g. 
```
const RELOAD_DATA_INTERVAL_MS = 1000 * 60; // 60s
```

src/GameData.ts
```diff
@@ -0,0 +1,240 @@
...
+                });
+            });
+
+        return returnData;
+    }
+
+    public async  loadPerkData(): Promise<PerkData[]> {
+        const returnData: PerkData[] = [];
+
+        const perkDataUrl = "https://raw.communitydragon.org/pbe/plugins/rcp-be-lol-game-data/global/default/v1/perks.json";
```
idunnololz: Make with an actual const pls. If this url might be used in multiple files, make it a global const or static class, or w.e is the standard in js.

src/GameData.ts
```diff
@@ -0,0 +1,240 @@
...
+        }));
+
+        const skinDataUrl = "https://raw.communitydragon.org/pbe/plugins/rcp-be-lol-game-data/global/default/v1/skins.json";
+        const skins = await (await fetch(skinDataUrl)).json() as SkinData[];
+
+        (champions as ChampionData[]).map(c => c.id)
+            .filter(id => id > 0)
+            .map(i => i.toString())
+            .forEach(id => {
+                const skinKeys = Object.keys(skins).filter(k => k.startsWith(id)).filter(k => (k.length - id.length) === 3).map(k => +k);
```
idunnololz: "Hey guys! I'm cool cause I write 1 liners".
```
const skinKeys = Object.keys(skins)
    .filter(k => k.startsWith(id))
    .filter(k => (k.length - id.length) === 3)
    .map(k => Number(k));
```

src/GameData.ts
stelar7: We have setting files for this.

src/GameData.ts
stelar7: Yes i am.

src/GameData.ts
```diff
@@ -0,0 +1,240 @@
...
+
+        return "```" + JSON.stringify(result[0], null, 4) + "```";
+    }
+
+    public async findPerk(search: string) {
+        if (!search || search === undefined || search == null || search.length === 0) {
+            return "There are currently ${perkData.length} perks on the PBE server!";
+        }
+
+        const result = this.perkData.filter(c => Math.max(levenshteinDistance(search, c.id.toString()), levenshteinDistance(search, c.name), levenshteinDistance(search, c.shortDesc)) < 3);
```
idunnololz: oh look. Another one liner!

stelar7: Outdated

src/GameData.ts
```diff
@@ -0,0 +1,265 @@
...
+            ]) < 1);
+
+        if (result.length === 0) {
+            return "Unable to find any data for that search term.";
+        }
+
+        if (result.length > 1) {
+            let response = "Too many results returned for that search, please try one of the options below:\n";
+            result.forEach(r => response += `\`${r}\`, `);
+            response = response.slice(0, -2);
```
Querijn: What does this exactly do?

src/GameData.ts
Querijn: I would agree, show up to 3 and any more you would omit with a message like `..x more results found.`

src/GameData.ts
Querijn: I agree though, handle both.

Querijn: Here's why:
image
image

src/GameData.ts
Querijn: Remember that if you add `rune` as an allowed argument to add it here too.

src/GameData.ts
```diff
@@ -0,0 +1,265 @@
...
+                levenshteinDistance(search, c.categories.toString()),
+            ]) < 1);
+
+        if (result.length === 0) {
+            return "Unable to find any data for that search term.";
+        }
+
+        if (result.length > 3) {
+            let response = "Too many results returned for that search, please try one of the options below:\n";
+            result.forEach(r => response += `\`${r.name}\`, `);
```
Querijn: This is a foreach loop through potentially hundreds of skins. We only take 3.

src/GameData.ts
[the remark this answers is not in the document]
Querijn: Or a setting

src/GameData.ts
```diff
@@ -0,0 +1,265 @@
...
+        this.champData = await this.loadChampionData();
+        this.perkData = await this.loadPerkData();
+        this.itemData = await this.loadItemData();
+        console.log("Game data loaded!");
+    }
+
+    public async loadChampionData(): Promise<ChampionData[]> {
+        const returnData: ChampionData[] = [];
+
+        const championDataUrl = "https://raw.communitydragon.org/latest/plugins/rcp-be-lol-game-data/global/default/v1/champion-summary.json";
```
Querijn: Can you make this a setting

src/GameData.ts
```diff
@@ -0,0 +1,265 @@
...
+                    const data = skins[key];
+                    returnData.filter(r => r.id === +id)[0].skins.push({ ...data });
+                });
+            });
+
+        return returnData;
+    }
+
+    public async loadPerkData(): Promise<PerkData[]> {
+        const perkDataUrl = "https://raw.communitydragon.org/latest/plugins/rcp-be-lol-game-data/global/default/v1/perks.json";
```
Querijn: Setting?

src/GameData.ts
stelar7: it removes the trailig commas, its getting replaces later when i go over molenz changes

src/GameData.ts
stelar7: Because `result[0]` contains all the fields from the json data (even with the ItemData[] cast).
The resulting json has over 2000 chars for most (if not all) items, so discord cant send it

src/LevenshteinDistance.ts
```diff
@@ -43,3 +43,10 @@ export default function levenshteinDistance(a: string, b: string): number {
 
     return v1[b.length];
 }
+
+/**
+ * Counts the substitutions needed to transform a into b for each element in b, and returns the lowest matching score
+ */
+export function levenshteinDistanceArray(a: string, b: string[]): number {
+    return b.map(x => levenshteinDistance(a, x)).sort((x, y) => x > y ? 1 : -1)[0];
```
molenzwiebel: `Math.min(...b.map(x => levenshteinDistance(a, x)))`?

src/GameData.ts
```diff
@@ -0,0 +1,262 @@
...
+                result = this.findPerk(args.slice(1).join(" "));
+                break;
+            }
+            case "champion": {
+                result = this.findChampion(args.slice(1).join(" "));
+                break;
+            }
+        }
+
+        if (typeof (result) === "string") {
```
molenzwiebel: Remove brackets

src/GameData.ts
```diff
@@ -0,0 +1,262 @@
...
+            message.channel.send(response);
+            return;
+        }
+
+        if (!supportedTypes.includes(args[0])) {
+            message.channel.send(`I'm sorry. I'm unable to parse the category \`${args[0]}\` at this moment. If you think it should be added, please contact a guru.`);
+            return;
+        }
+
+        let result;
```
molenzwiebel: `let result: string | any[]`

src/GameData.ts
```diff
@@ -0,0 +1,262 @@
...
+                break;
+            }
+        }
+
+        if (typeof (result) === "string") {
+            message.channel.send(result);
+            return;
+        }
+
+        (result as string[]).forEach(x => message.channel.send("```" + JSON.stringify(x, null, 4) + "```"));
```
molenzwiebel: No need to cast here if you add the type annotation

src/GameData.ts
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    public findItem(search: string): string | {} {
```
molenzwiebel: `string | any[]`

src/GameData.ts
```diff
@@ -0,0 +1,262 @@
...
+            id: x.value.id,
+            name: x.value.name,
+            combineCost: x.value.price,
+            cost: x.value.priceTotal,
+            from: x.value.from,
+            to: x.value.to,
+        }));
+    }
+
+    public findPerk(search: string): string | {} {
```
molenzwiebel: `string | any[]`

src/GameData.ts
```diff
@@ -0,0 +1,262 @@
...
+        }
+
+        return result.slice(0, this.sharedSettings.lookup.returnedEntryCount).map(x => ({
+            id: x.value.id,
+            name: x.value.name,
+            endOfGameStatDescs: x.value.endOfGameStatDescs,
+        }));
+    }
+
+    public findChampion(search: string): string | {} {
```
molenzwiebel: `string | any[]`

src/GameData.ts
Querijn: Is this resolved @stelar7?

src/GameData.ts
```diff
@@ -0,0 +1,262 @@
...
+            idscore: Math.min(
+                levenshteinDistance(search, c.id.toString()),
+            ),
+        })).filter(x => x.textscore < this.sharedSettings.lookup.textConfidence || x.idscore < this.sharedSettings.lookup.numberConfidence);
+
+        if (result.length === 0) {
+            return "Unable to find any data for that search term.";
+        }
+
+        result.sort((a, b) => Math.min(a.textscore, a.idscore) > Math.min(b.textscore, b.idscore) ? 1 : -1);
```
Querijn: This can compare idscore vs textscore. Is that intended?

src/GameData.ts
```diff
@@ -0,0 +1,262 @@
...
+            value: c,
+            textscore: Math.min(
+                levenshteinDistance(search, c.name),
+            ),
+            idscore: Math.min(
+                levenshteinDistance(search, c.id.toString()),
+            ),
+        })).filter(x => x.textscore < this.sharedSettings.lookup.textConfidence || x.idscore < this.sharedSettings.lookup.numberConfidence);
+
+        if (result.length === 0) {
```
Querijn: This function seems to work unexpected: (Ignore the savagery request)
image

If the request is not exact, it gives no results.

Expected is something alike this:
image

src/GameData.ts
```diff
@@ -0,0 +1,262 @@
...
+            return "Unable to find any data for that search term.";
+        }
+
+        result.sort((a, b) => Math.min(a.textscore, a.idscore) > Math.min(b.textscore, b.idscore) ? 1 : -1);
+
+        if (result.length > this.sharedSettings.lookup.returnedEntryCount) {
```
Querijn: Don't you agree Annie would be on top of this list?
image

Querijn: There's a bug with this function: 
image

I think any ID returns Vi as well.

src/GameData.ts
stelar7: Yes. It should pick the lowest score, no matter where it was from.

src/GameData.ts
stelar7: The text match is set to match up-to 2 different characters, so it wont match the other ones unless its changed to a larger number. Might want to separate it for champs / runes / items, since champs have relatively shorter names.

src/GameData.ts
stelar7: Thats a result of the textsearch having a value of 2, since vi only has 2 chars, she always matches.. 
I'll make some changes to it when i come home to fix it.

src/GameData.ts
```diff
@@ -0,0 +1,320 @@
...
+    private itemData: ItemData[];
+
+    private bot: Discord.Client;
+    private sharedSettings: SharedSettings;
+
+    public constructor(bot: Discord.Client, settings: SharedSettings) {
+        this.bot = bot;
+        this.sharedSettings = settings;
+
+        bot.on("ready", async () => {
```
molenzwiebel: why `async` tho

src/GameData.ts
```diff
@@ -0,0 +1,320 @@
...
+            this.reloadData();
+            setInterval(this.reloadData, this.sharedSettings.lookup.refreshTimeout);
```
molenzwiebel: you gotta `bind` or this won't work

src/GameData.ts
```diff
@@ -0,0 +1,320 @@
...
+                break;
+            }
+        }
+
+        if (typeof result === "string") {
+            message.channel.send(result);
+            return;
+        }
+
+        result.forEach(x => message.channel.send("```" + JSON.stringify(x, (k, v) => { if (v !== null) return v; }, 4) + "```"));
```
molenzwiebel: `(k, v) => v !== null ? v : undefined` (or `void 0`)

molenzwiebel: Also add a comment saying that you're using this to remove keys with `null` as value from the json representation. The second argument is rarely used so people may not know what it does.

src/GameData.ts
```diff
@@ -0,0 +1,320 @@
...
+
+        if (nameA.startsWith(search) && !nameB.startsWith(search)) return -1;
+        if (nameB.startsWith(search) && !nameA.startsWith(search)) return 1;
+
+        if (nameA.includes(search) && !nameB.includes(search)) return -1;
+        if (nameB.includes(search) && !nameA.includes(search)) return 1;
+
+        if (a.item.key && b.item.key) {
+            const keyA = a.item.key!;
+            const keyB = b.item.key!;
```
molenzwiebel: You shouldn't need the `!` here, typescript infers it cannot be null/undefined.

src/GameData.ts
```diff
@@ -0,0 +1,320 @@
...
+    id: number;
+    name: string;
+    categories: string[];
+    price: number;
+    priceTotal: number;
+    iconPath: number;
+    from: string[];
+    to: string[];
+}
+interface SearchType {
```
molenzwiebel: This is a bad name. SearchType makes me think it is an enum of different sorting strategies, not a container that has items and a score.

src/GameData.ts
```diff
@@ -0,0 +1,320 @@
...
+
+        return searchResult.map(x => x.item as ItemData)
+            .map(x => ({
+                id: x.id,
+                name: x.name,
+                combineCost: x.price,
+                cost: x.priceTotal,
+                from: x.from.length > 0 ? x.from : null,
+                to: x.to.length > 0 ? x.to : null,
+            })).slice(0, 1);
```
molenzwiebel: Should probably make `from` and `to` empty arrays, since they will be removed if they are null. IMO it makes more sense to show them as empty than to not show them entirely, but feel free to correct me.

src/GameData.ts
```diff
@@ -0,0 +1,320 @@
...
+            return `There are currently ${this.itemData.length} items in my lookup data!`;
+        }
+
+        let searchResult: SearchType[];
+        if (search.match(/^\d+$/)) {
+            searchResult = this.itemData.map(i => ({ item: i, score: levenshteinDistance(search, i.id.toString()) }));
+        } else {
+            searchResult = this.itemData.map(i => ({ item: i, score: levenshteinDistance(search, i.name.toLowerCase()) }));
+        }
+
```
molenzwiebel: Can probably do 

```js
const searchResult = this.itemData.map(i => ({
	item: i,
	score: levenshteinDistance(search, search.match(/^\d+$/) ? i.id.toString() : i.name.toLowerCase())
}));
```

src/GameData.ts
```diff
@@ -0,0 +1,320 @@
...
+                return "Unable to find any data for that search term.";
+            }
+        }
+
+        return searchResult.map(x => x.item as PerkData)
+            .map(x => ({
+                id: x.id,
+                name: x.name,
+                endOfGameStatDescs: x.endOfGameStatDescs,
+            })).slice(0, 1);
```
molenzwiebel: Why the `slice` call? Now you're returning an array with a single element instead of an object. That just causes extra `[]`s in the response.

PR #188, opened 2019-04-07
Add champ as alias for champion, use JSON highlighting

molenzwiebel:


src/GameData.ts
```diff
@@ -146,7 +146,8 @@ export default class GameData {
                 result = this.findPerk(searchTerm);
                 break;
             }
-            case "champion": {
+            case "champion": // fall through
+            case "champ": {
```
stelar7: isnt there a fancy ES2018 thing we can do to merge the cases? like `case: ["champ", "champion"]`

In other langs you can do things like this, so i assume JS has something similar?
```

   final String numericString = switch (integer)  
      {  
         case 0 -> "zero";  
         case 1, 3, 5, 7, 9 -> "odd";  
         case 2, 4, 6, 8, 10 -> "even";  
         default -> "N/A";  
      };  
```

molenzwiebel: Nope. Fallthrough is the best way.

src/GameData.ts
```diff
@@ -157,7 +158,7 @@ export default class GameData {
             return;
         }
 
-        message.channel.send("```" + JSON.stringify(result, null, 4) + "```");
+        message.channel.send("```json\n" + JSON.stringify(result, null, 4) + "```");
```
stelar7: message.channel.send("```" + JSON.stringify(result, null, 4) + "```", {code:"json"});

molenzwiebel: Then you'd have to leave out the backticks too. `message.channel.send(JSON.stringify(result, null, 4), {code:"json"})`. Both work, but IMO the first one is more clear what is actually sent.

Querijn:
It didn't work, there's an array at the start that was missing the "champ" keyword.

PR #189, opened 2019-04-08
Commands are case insensitive and no crashes on incorrect library request

### LINKED ISSUE #185: using !libs with a bad language results in botty crashing due to an empty message being sent

### LINKED ISSUE #186: commands should be case insensitive

Djinnes:
Closes #186 
Closes #185 
Ignore first commit, forgot that we don't want text after that made to lower case.

### DIFF
--- a/src/RiotAPILibraries.ts
+++ b/src/RiotAPILibraries.ts
@@ -122,7 +122,7 @@ export default class RiotAPILibraries {
     }
 
     private async initList() {
-        
+
         const response = await fetch(this.settings.riotApiLibraries.baseURL, this.fetchSettings);
         if (response.status !== 200) {
             console.error(this.settings.riotApiLibraries.githubErrorList + response.status);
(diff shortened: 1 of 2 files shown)


Querijn:
Seems fine, should work as indicated. I'll test this tonight, but I don't think there are any problems with this.

stelar7:
This assumes all aliases are lowercase in the commands file. might want to also make those lowercase on startup

Djinnes:
Hmm, I also did #185, is it good practise to go back to the current master, make the change and do a separate pull request?

src/RiotAPILibraries.ts
```diff
@@ -181,7 +181,7 @@ export default class RiotAPILibraries {
 
         // Check if alias
         for (const [key, values] of Object.entries(this.settings.riotApiLibraries.aliases)) {
-            if (values.find(self => self.toLowerCase() === language.toLowerCase())) {
+            if (values.find(self => self.toLowerCase() === language)) {
```
stelar7: Whats the reason behind removing the .toLowerCase() here?

stelar7:
Ideally each PR should take on one issue, and go from the master branch, but its fine usually

src/RiotAPILibraries.ts
Djinnes: https://github.com/Querijn/BottyMcBotface/blob/0f6cd50bee4775dfefe180a064125f1d1ef70cb2/src/RiotAPILibraries.ts#L85

It's already made into lowercase when it's selected.

Issue #195, opened 2019-04-24
Change !lookup to return an embed


stelar7:
By putting the data inside an embed we can also return the image of the lookup (item/champ/perk)

The content field could be optional
Paste the below into https://leovoel.github.io/embed-visualizer/ for an example
```
{
  "content": "```{\r\n  \"id\": 4026,\r\n  \"name\": \"Mariner's Vengeance\",\r\n  \"combineCost\": 100,\r\n  \"cost\": 2800,\r\n  \"from\": [\r\n    \"B. F. Sword\",\r\n    \"Chain Vest\",\r\n    \"Stopwatch\"\r\n  ],\r\n  \"to\": [\r\n    \r\n  ]\r\n}```",
  "embed": {
    "thumbnail": {
      "url": "https://raw.communitydragon.org/pbe/plugins/rcp-be-lol-game-data/global/default/data/items/icons2d/4026_mariners_vengeance.png"
    },
    "fields": [
      {
        "name": "id",
        "value": "4026",
        "inline": true
      },
      {
        "name": "name",
        "value": "Mariner's Vengeance",
        "inline": true
      },
      {
        "name": "cost",
        "value": "2800",
        "inline": true
      },
      {
        "name": "combineCost",
        "value": "100",
        "inline": true
      },
      {
        "name": "from",
        "value": "B. F. Sword \nChain Vest \nStopwatch",
        "inline": true
      },
      {
        "name": "to",
        "value": "​",
        "inline": true
      }
    ]
  }
}
```

Querijn:
@Krissalyce 
Yes, it does have a way. 

Should all be in here: https://github.com/Querijn/BottyMcBotface/blob/master/src/GameData.ts

PR #205, opened 2019-08-05
Change !lookup to return an embed (Part 1)

Krissalyce:
Addresses #195 , initial steps to returning an embed. Linking back to the champion/perk/item.json not yet implemented.


Krissalyce:
Ran into a new error when testing this afternoon, so going to close this PR until I can fix it.

PR #222, opened 2019-11-21
Nerf !status

Token07:


Querijn:
I don't like it. If you want to show someone else the status, this kind of falls apart. If we do some kind of action, I'd say it would edit an old post and post a link to it instead.

RayZz-:
i have a suggestion where there can be a channel which gets updated by botty when ever a change to the status happens or every so often

PR #223, opened 2019-11-22
Add ability to answer questions (resolves #200)

### LINKED ISSUE #200: Add !answer
Add `!answer` as a component of `!ask`, this way questions can be answered without opening #office-hours

Token07:


Querijn:
Seems to be pretty prone to trolling.. Don't think I'd like to check it in as-is. They're usually questions directly at Riot. Usually Gurus answer in the chat itself if it's easy to, and close the question. What's the usecase for this?

src/OfficeHours.ts
```diff
@@ -145,6 +149,28 @@ export default class OfficeHours {
...
+        const officehoursChannel = this.guild.channels.find("name", "office-hours");
+        const question = this.data.questions.find(q => q.uuid === id);
+
+        if (question) {
+            if (officehoursChannel instanceof Discord.TextChannel) {
+                const answer = joinArguments(args, separators.slice(1), 1);
+                officehoursChannel.send(`${message.author.toString()} answered a question asked by ${question.authorMention} \n ${question.question}\n\nAnswer: ${answer}`);
+                this.data.questions = this.data.questions.filter(q => q.uuid !== id);
+
+                const moderatorChannel = this.guild.channels.find("name", "moderators");
```
Querijn: This is no longer correct. It's gurus now, and it's in the settings.

src/OfficeHours.ts
```diff
@@ -145,6 +149,28 @@ export default class OfficeHours {
         }
     }
 
+    public onQuestionAnswer(message: Discord.Message, isAdmin: boolean, command: string, args: string[], separators: string[]) {
+        const id = +args[0];
+        const officehoursChannel = this.guild.channels.find("name", "office-hours");
```
Querijn: Is there a setting for this guild channel name? If not, let's add it.

PR #234, opened 2020-02-07
change game version checking so that it uses the json from the website

stelar7:


src/VersionChecker.ts
```diff
@@ -157,7 +194,7 @@ export default class VersionChecker {
 
     private async onUpdate() {
         await this.updateDataDragonVersion();
-        await this.updateGameVersion();
+        await this.updateGameVersion2020();
```
Querijn: why

stelar7: cus you might want to swap back some time :shrug:

Querijn: We have version control for those cases

PR #244, opened 2020-05-17
Update notes for ddragon & dev portal

bangingheads:
Change ddragon urls to zip, edit dev portal to new urls, add missing notes, remove notes that urls no longer exist


Querijn:
Can you reset the formatting? It makes the diff unreadable.

PR #262, opened 2022-04-01
Elliejs !libs patch

elliejs:
- !libs only works in dev channels, with error and suggestion message in the rest
- replies with libs in a context aware manner to the channel
- all parameters are hoisted to the settings file


elliejs:
lmk other comments for review. ignoring this until there's activity 😄

pseudonym117:
This doesnt really make sense to merge without also updating the [library list](https://github.com/WxWatch/riot-api-libraries) with these new required tags for libraries that support each different game.... will just cause more confusion

elliejs:
What if I just add V4 (the only tag being listened to currently), to every value list of the context map? This would mean that all libraries that already show up will continue to show up, and libraries using updated and more usefully specific tags will show up in relevant channels.

PR #273, opened 2023-03-19
Ignore "Onboarded" role

Token07:
Onboarded is automatically given when joining server now. Ignore for the purposes of SpamKiller check.

### DIFF
--- a/src/SpamKiller.ts
+++ b/src/SpamKiller.ts
@@ -150,7 +150,7 @@ export default class SpamKiller {
         if (!member)
             throw new Error(`Unable to find member that wrote the message '${message.content}' (${message.author.username})`);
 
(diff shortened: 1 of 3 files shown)


src/SpamKiller.ts
```diff
@@ -150,7 +150,7 @@ export default class SpamKiller {
         if (!member)
             throw new Error(`Unable to find member that wrote the message '${message.content}' (${message.author.username})`);
 
-        if (member.roles.cache.size > 1) { // Only act on people without roles
+        if (member.roles.cache.filter(r => r.id != "711261336285347939").size > 1) { // Only act on people without roles
```
Querijn, line 153: Would recommend moving this ID to the /settings/shared_settings.json file

Querijn (changes requested):

PR #275, opened 2023-03-20
Fix check for allowed URLs

Token07:
Previously it only matched if the URL was riotgames.com or leagueoflegends.com, but did not allow subdomains.

### DIFF
--- a/src/SpamKiller.ts
+++ b/src/SpamKiller.ts
@@ -134,9 +134,9 @@ export default class SpamKiller {
 
         const d = url.parse(urlString);
         const hostname = d.hostname || "";
-        if (this.sharedSettings.spam.allowedUrls.findIndex(u => hostname.endsWith(u)) !== -1)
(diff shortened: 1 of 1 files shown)


src/SpamKiller.ts
```diff
@@ -134,7 +134,7 @@ export default class SpamKiller {
 
         const d = url.parse(urlString);
         const hostname = d.hostname || "";
-        if (this.sharedSettings.spam.allowedUrls.findIndex(u => u.endsWith(hostname)) !== -1)
+        if (this.sharedSettings.spam.allowedUrls.findIndex(u => hostname.endsWith(u)) !== -1)
```
Querijn, line 137: This allows hackleagueoflegends.com

Querijn (approved):

Querijn (changes requested):
Please resolve the comments

src/SpamKiller.ts
Token07: Should be fixed now.

PR #276, opened 2023-03-20
botty tag context aware channel responses (non behaviour-breaking)

elliejs:


elliejs:
Untested changes since I don't actually know how to host a discord bot lol. But the linter seems to think everything is fine

Querijn:
@Token07 Could you test these changes?

src/RiotAPILibraries.ts
```diff
@@ -217,34 +211,19 @@ export default class RiotAPILibraries {
...
-            }
-            libraryDescriptions = (await this.getLibrariesForLanguage(language, requiredTags))
-                .sort((a, b) => b.stars - a.stars); // Sort by stars
-        }
-        catch (e) {
-            message.channel.send(e.message);
-            return;
+        let requiredTags = this.allTagOptions;
+        
+        if(message.channel.type == 'GUILD_TEXT') { // if this is the server text channel
```
Token07, line 220: error TS2367: This condition will always return 'false' since the types '"text" | "dm" | "news"' and '"GUILD_TEXT"' have no overlap.

src/RiotAPILibraries.ts
```diff
@@ -157,7 +157,24 @@ export default class RiotAPILibraries {
     }
 
     private async getList(message: Discord.Message) {
-        const reply = this.settings.riotApiLibraries.languageList.replace("{languages}", "`" + this.languageList.join(", ") + "`");
+        let applicableLangs = Array.from(this.languageMap.keys());
+
+        if(message.channel.type == 'GUILD_TEXT') { // if this is the server text channel
```
Token07, line 162: error TS2367: This condition will always return 'false' since the types '"text" | "dm" | "news"' and '"GUILD_TEXT"' have no overlap.

src/RiotAPILibraries.ts
```diff
@@ -157,7 +157,24 @@ export default class RiotAPILibraries {
...
+            let requiredTags = this.settings.riotApiLibraries.requiredTagContextMap.get(message.channel.name) || []; //get the reqiured tags from settings.
```
Token07, line 163: error TS2339: Property 'name' does not exist on type 'never'.

Token07 (changes requested):
These are the errors I get when attempting to build.

src/RiotAPILibraries.ts
elliejs: fixed from "GUILD_TEXT" to "text", which should be the right keyword for discord.js..... though that's definitely not what the docs suggest:
Message.channel is of type [TextBasedChannels](https://discord.js.org/#/docs/discord.js/main/typedef/TextBasedChannels) which eventually as a aggregate type has ChannelType which contains a huge enum. None of this is a string union type but..... I guess I'll listen to the linter and not the docs?

elliejs:
@Token07 I think the branch is re-ready for a build. Sorry again for doing this by ear, I have no experience with discord bots so hosting my own build system is not something I am comfortable getting set up.

src/RiotAPILibraries.ts
```diff
@@ -68,7 +69,8 @@ export default class RiotAPILibraries {
                 "Content-Type": "application/json",
             },
         };
-        this.allTagOptions = Array.from(this.settings.riotApiLibraries.requiredTagContextMap.values()).flat();
+        this.allTagOptions =([] as string[]).concat(...this.settings.riotApiLibraries.requiredTagContextMap.values());
```
Token07, line 72: I get: `this.settings.riotApiLibraries.requiredTagContextMap.values is not a function`
I ended up changing this to Object.values(this.settings.riotApiLibraries.requiredTagContextMap) to get a little further.

Token07 (changes requested):
Got a few more errors. Can build with tsc now, but doesn't run.

src/RiotAPILibraries.ts
elliejs: right ok this is because the map isn't constructed properly from the json shared_settings. I need to fix the json object to map conversion; or leave it as an object but I'm not sure what the typing looks like on that

Token07 (changes requested):

src/RiotAPILibraries.ts
```diff
@@ -99,7 +99,7 @@ export default class RiotAPILibraries {
         const libraryResponse = await fetch(json.download_url);
         const libraryInfo: APILibraryStruct = await libraryResponse.json();
 
-        if (!libraryInfo.tags || libraryInfo.tags.indexOf("v4") === -1) {
+        if (!libraryInfo.tags.some(tag => self.allTagOptions.includes(tag))) {
```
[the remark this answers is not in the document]
elliejs: yep that's a facepalm alright. self -> this.... smh my head

Token07:
Some more errors:
```
Unable to fetch library data for language java: TypeError: Cannot read properties of undefined (reading 'some')
Unable to fetch library data for language javascript: TypeError: Cannot read properties of undefined (reading 'some')
Unable to fetch library data for language objective-c: TypeError: Cannot read properties of undefined (reading 'some')
Unable to fetch library data for language perl: TypeError: Cannot read properties of undefined (reading 'some')
Unable to fetch library data for language php: TypeError: Cannot read properties of undefined (reading 'some')
Unable to fetch library data for language python: TypeError: Cannot read properties of undefined (reading 'some')
Unable to fetch library data for language ruby: TypeError: Cannot read properties of undefined (reading 'some')
Unable to fetch library data for language typescript: TypeError: Cannot read properties of undefined (reading 'some')
Riot API library languages updated: cpp, elixir, go, julia, rust, swift
```

elliejs:
yes I do that would avoid this trainwreck.

src/RiotAPILibraries.ts
```diff
@@ -99,7 +99,7 @@ export default class RiotAPILibraries {
         const libraryResponse = await fetch(json.download_url);
         const libraryInfo: APILibraryStruct = await libraryResponse.json();
 
-        if (!libraryInfo.tags.some(tag => this.allTagOptions.includes(tag))) {
+        if (!libraryInfo.tags || libraryInfo.tags.some(tag => this.allTagOptions.includes(tag)) === -1) {
```
Token07, line 102: This should be false, tsc is complaining about it.
`src/RiotAPILibraries.ts:102:34 - error TS2367: This condition will always return 'false' since the types 'boolean' and 'number' have no overlap.`

elliejs: oh good lord, I'm just not understanding what typescript wants me to do with these maybe-null/undefined (and those are apparently different types) values

elliejs: ok, TS was being more reasonable than I thought. yet another commit

Token07:
It runs but:
!libs with no argument does not list any languages outside of a dm

In DM and channel context !libs [js/cpp/etc] returns an embed with no libraries in it. i.e. All it has is "List of libraries for java:"

Token07:
If you need help setting up a copy of Botty lmk. I'm Boaty McBoatface on Riot API discord.

Token07:
Looking better 
```
I have libraries for one of the following languages: c-sharp, c-sharp, c-sharp, cpp, elixir, go, java, java, java, javascript, javascript, javascript, javascript, javascript, javascript, javascript, javascript, javascript, julia, php, php, php, python, python, python, rust, rust, swift, swift, typescript, typescript, typescript
Choose one you like and reply with !libs <your language>.```

elliejs:
even if these work, dont pull; i've got refactoring on nullish stuff to do

Querijn:
How's that coming along? Can we close this and you open a new ticket when you've finished that up?

PR #277, opened 2023-03-22
Correct ID for Onboarded role and add a couple more "banned" words

Token07:


Querijn:
You sure about the word "help"?

PR #282, opened 2023-04-10
Add kick and ban commands to Botty

DarkIntaqt:
Based on [this request](https://discord.com/channels/187652476080488449/187652476080488449/1094657023615959062), Botty can now !kick or !ban members. 
The kick and ban commands supports multiple users, tagged or as a raw snowflake. Additionally, a .note can be added as a quick ban message. 

## Example: 
- `!ban @user 1234snowflake56789 .we-mad` 
- `!ban @user` => no message is ok
- `!kick 1234snowflake56789 .shipment you got kicked with a shipment` => text after the note is also totally fine
- `!kick @user no note is also good to go`
The user will be notified if a ban message was set

## Security:
- Requires an admin role
- Can't kick/ban users with an admin role

## Features that could be added but I wasn't sure so lets discuss:
- Delete the messages of a banned user (within last x days / weeks etc)
- Delete the `!kick` || `!ban` command in a non-guru-channel

### DIFF
--- a/src/Admin.ts
+++ b/src/Admin.ts
@@ -5,6 +5,7 @@ import { clearTimeout, setTimeout } from "timers";
 import Discord = require("discord.js");
 import prettyMs = require("pretty-ms");
 import joinArguments from "./JoinArguments";
+import Info from "./Info";
 
 class TicketData {
 
@@ -57,10 +58,12 @@ export default class Admin {
     private data: AdminData;
     private muteRole: Discord.Role;
     private muteTimers: { [id: string]: NodeJS.Timer } = {};
+    private notes: Info;
 
-    constructor(bot: Discord.Client, sharedSettings: SharedSettings, dataFile: string) {
+    constructor(bot: Discord.Client, sharedSettings: SharedSettings, dataFile: string, notes: Info) {
         console.log("Requested Admin extension..");
         this.bot = bot;
+        this.notes = notes;
 
         this.sharedSettings = sharedSettings;
         this.data = fileBackedObject<AdminData>(dataFile);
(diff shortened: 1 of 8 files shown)


src/Admin.ts
```diff
@@ -120,6 +123,183 @@ export default class Admin {
...
+                });
+            }
+
+            let catched = false;
+            member.kick((reason.length > 0 ? this.shorten(reason) : "No reason") + " -" + message.author.username).catch(e => {
+                catched = true;
+                this.replySecretMessage(message, `Failed to kick ${member}: ${e}`);
+            })
+
+            if (!catched) {
```
Derpthemeus, line 288: Does this work? `catched` is set asynchronously

src/Admin.ts
```diff
@@ -120,6 +123,183 @@ export default class Admin {
...
+                });
+            }
+
+            let catched = false;
+            member.ban({ reason: (reason.length > 0 ? this.shorten(reason) : "No reason") + " -" + message.author.username }).catch(e => {
+                catched = true;
+                this.replySecretMessage(message, `Failed to ban ${member}: ${e}`);
+            })
+
+            if (!catched) {
```
Derpthemeus, line 240: Does this work? `catched` is set asynchronously

Derpthemeus:
> Delete the !kick || !ban command in a non-guru-channel

I'm a big fan of adding this support, no need to clutter channels with kick/ban history. There are 3 channels we'd want to keep the commands in, and I'm not sure if they're all in config yet: `#gurus` (865107613992616007), `#admins-only` (866816996900864041), and `#botty-test-channel` (429367396420157460)

DarkIntaqt:
> > Delete the !kick || !ban command in a non-guru-channel
> 
> I'm a big fan of adding this support, no need to clutter channels with kick/ban history. There are 3 channels we'd want to keep the commands in, and I'm not sure if they're all in config yet: `#gurus` (865107613992616007), `#admins-only` (866816996900864041), and `#botty-test-channel` (429367396420157460)

I'm also a big fan of this feature. Currently, they are not in the config (except gurus), so I will add them to config.admin.keepAdminCommands[].

src/Admin.ts
```diff
@@ -120,6 +123,183 @@ export default class Admin {
...
+
+            if (reason.length > 0) {
+                note = "The user received their ban message. ";
+                await member.send({ embed: this.getDMEmbed(reason, "banned", message) }).catch(e => {
+                    note = "The user did not receive their ban message. "
+                });
+            }
+
+            let catched = false;
+            member.ban({ reason: (reason.length > 0 ? this.shorten(reason) : "No reason") + " -" + message.author.username }).catch(e => {
```
Token07:
```js
.then(() => {
if (reason.length > 0) {
    this.replySecretMessage(message, `${member} was banned by ${message.author.username} due to "${reason}". ${note}`);
...

}).catch(e => {
...
```
Would probably look better

DarkIntaqt: I refactored the design a little bit earlier, that's why I can't 1:1 use this code. 
I hope the new commit still matches your idea (LL 259 - 267)

PR #286, opened 2023-04-18
Use embed for checkForLinks

Token07:


src/SpamKiller.ts
```diff
@@ -155,7 +155,13 @@ export default class SpamKiller {
         if (this.sharedSettings.spam.allowedUrls.findIndex(u => hostname.endsWith(u) &&
         (hostname.replace(u, "").endsWith(".") || hostname.replace(u, "").length === 0)) !== -1) // Only allow matching base domain (zero length after replace) and subdomains (ends with ".")
             return;
-        this.addViolatingMessage(message, `Hey, ${message.author}, we require users to verify that they are human before they are allowed to post a link. If you are a human, react with :+1: to this message to gain link privileges. If you are a bot, please go spam somewhere else. 👍`);
+
+        const embed = new Discord.MessageEmbed()
+            .setTitle("Robot Check")
+            .setColor(0xffff00)
+            .setThumbnail("https://upload.wikimedia.org/wikipedia/commons/thumb/3/39/Nuvola_apps_important_yellow2.svg/288px-Nuvola_apps_important_yellow2.svg.png")
+            .setDescription("We require users to verify that they are human before they are allowed to post a link. If you are a human, react with :+1: to this message to gain link privileges. If you are a bot, please go spam somewhere else. 👍");
```
renja-grotemeyer, line 163: Why are you using `:+1:` here and later 👍 ?

src/SpamKiller.ts
```diff
@@ -155,7 +155,13 @@ export default class SpamKiller {
         if (this.sharedSettings.spam.allowedUrls.findIndex(u => hostname.endsWith(u) &&
         (hostname.replace(u, "").endsWith(".") || hostname.replace(u, "").length === 0)) !== -1) // Only allow matching base domain (zero length after replace) and subdomains (ends with ".")
             return;
-        this.addViolatingMessage(message, `Hey, ${message.author}, we require users to verify that they are human before they are allowed to post a link. If you are a human, react with :+1: to this message to gain link privileges. If you are a bot, please go spam somewhere else. 👍`);
+
+        const embed = new Discord.MessageEmbed()
+            .setTitle("Robot Check")
+            .setColor(0xffff00)
```
renja-grotemeyer, line 161: I think this plain yellow looks a bit retro.

Maybe `#FFCC00` would be a better fit.
https://coolors.co/ffff00-ffcc00

renja-grotemeyer: The png also looks a bit old.

src/SpamKiller.ts
Token07: It was written like that originally, I just cut and paste it

src/SpamKiller.ts
Token07: I normally pick something out of the Nuvola set of icons, but I've updated the color and picture.

PR #288, opened 2023-05-07
Adding a command that shows how many peoply are banned on the server

renja-grotemeyer:
Since in the server settings you only get a list of banned people, but no counter, I added a command, that shows how many people are banned.

Since I have no local instance of Botty running, I have no idea if this command works, I just tried to recreate https://github.com/Querijn/BottyMcBotface/commit/b5a14a70c96181c7d4dc95b4103ab9d3e677fdbd

### DIFF
--- a/settings/command_list.json
+++ b/settings/command_list.json
@@ -237,5 +237,12 @@
             ],
             "description": "Shows a list of all Riot API endpoints."
         }
+    },
+    "banCount": {
+        "aliases": [
+            "ban_count",
+            "bancount"
+        ],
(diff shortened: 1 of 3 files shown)


Querijn:
You can see the people that were banned? That seems like something I'm not completely ok with, how do you do that?

renja-grotemeyer:
1. I can't.
2. The command would only display a number, no details.

But I could also move the command into Admin.ts

renja-grotemeyer:
I was trying to change the source branch before closing the PR, but it is not possible.

PR #289, opened 2023-05-12
Implemented sending the response as JSON file

renja-grotemeyer:
Since the webs server is not working
I made it so that the file is now just sent into the channel, no need to store it in the file system.
But I could also re add the file system and just send the file on top of storing the file.

### DIFF
--- a/src/ApiUrlInterpreter.ts
+++ b/src/ApiUrlInterpreter.ts
@@ -187,7 +187,7 @@ export default class ApiUrlInterpreter {
             return;
         }
 
-        const replyMessages = await message.channel.send(`Making a request to \`${path!.name}\``);
+        const replyMessages = await message.channel.send(`Making a request to ${path!.name}`);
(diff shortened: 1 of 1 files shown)


renja-grotemeyer:
grafik

src/ApiUrlInterpreter.ts
```diff
@@ -187,7 +187,7 @@ export default class ApiUrlInterpreter {
             return;
         }
 
-        const replyMessages = await message.channel.send(`Making a request to ${path!.name}`);
+        const replyMessages = await message.channel.send(`Making a request to \`${path!.name}\``);
```
bangingheads, line 190: This change seems very unnecessary.

renja-grotemeyer: Well yes, it was just for formatting:

FROM:
grafik


TO:
grafik

renja-grotemeyer: I reverted it

Token07 (approved):
LGTM

PR #292, opened 2023-05-22
Update guru command log channel

Token07:


Querijn (changes requested):
Content-wise this makes no sense. I would rename guruChannel as well.

Querijn:
Something got lost in translation here I think

What I meant fundamentally is that the variable is used to indicate the moderators channel. Changing the value to a non-moderators channel means that anything referencing has to correctly be changed to indicate that this value is in fact nor a moderators channel.

This PR has the issue that `this.adminChannel` now no longer refers to an admin channel, but the location where content changes appear. Admin.channel also wouldn't refer to the admin channel anymore, but rather the content-alerts channel. 

This could prove to be an brain-scratcher for future PR creators

I recommend either:
1. Changing all references to Admin.channel or Admin.adminChannel to logChannel and adminLogChannel respectively
2. Keeping the reference to the guruChannel the same and adding a variable for the log channel./

src/Admin.ts
```diff
@@ -81,14 +82,16 @@ export default class Admin {
...
         if (!adminChannel) {
             if (this.sharedSettings.botty.isProduction) {
                 console.error(`Admin: Unable to find moderators channel!`);
                 return;
             }
             else {
-                adminChannel = await guild.channels.create({name: this.sharedSettings.server.guruChannel, type: Discord.ChannelType.GuildText});
+                adminChannel = await guild.channels.create({name: this.sharedSettings.server.guruLogChannel, type: Discord.ChannelType.GuildText});
             }
+            adminLogChannel = guild.channels.cache.find(c => c.name === this.sharedSettings.server.guruLogChannel) || adminChannel;
```
Querijn, line 94: We only initialise the log channel if we can't find the admin channel?

Querijn (changes requested):

src/Admin.ts
Token07: Fixed

src/Admin.ts
```diff
@@ -90,6 +92,13 @@ export default class Admin {
                 adminChannel = await guild.channels.create({name: this.sharedSettings.server.guruChannel, type: Discord.ChannelType.GuildText});
             }
         }
+        if (!adminLogChannel) {
+            try {
+                adminLogChannel = await guild.channels.create({name: this.sharedSettings.server.guruLogChannel, type: Discord.ChannelType.GuildText});
```
Querijn, line 97: Needs `this.sharedSettings.botty.isProduction` check (see line 87)

Querijn (changes requested):

src/Admin.ts
Token07: Done

PR #295, opened 2023-06-21
Misc fixes

Token07:


src/Admin.ts
```diff
@@ -153,8 +153,9 @@ export default class Admin {
 
                 const id = (snowflake === false) ? argument.substring(2, argument.length - 1) : argument;
                 try {
-                    const member = await this.muteRole.guild.members.fetch(id);
-                    if (this.sharedSettings.commands.adminRoles.some(x => member.roles.cache.has(x))) {
+                    let member = this.muteRole.guild.members.cache.find(cachedMember => cachedMember.id === id);
+                    if (!member) member = await this.muteRole.guild.members.fetch(id);
+                    if (this.sharedSettings.commands.adminRoles.some(x => member!.roles.cache.has(x))) {
                         continue; // better not ban/kick admins here
                     }
```
Querijn, line 161: Can you explain what's going on here? Is there a possibility we can get rid of the forloop altogether?

src/Logger.ts
```diff
@@ -106,10 +106,10 @@ export default class Logger {
         this.oldLog(message, ...optionalParams);
 
         try {
-            let chunks = `[${(new Date()).toUTCString()}] Log: ${message.toString()}`.match(/.{1,2000}/g) || [message.toString()];
+            let chunks = `[${(new Date()).toUTCString()}] Log: ${message.toString()}`.match(/.{1,2000}/sg) || [message.toString()];
```
Querijn, line 109: What does this do?

Querijn (changes requested):

src/Logger.ts
Token07: Adding the s flag in allows `.` to match newlines

PR #305, opened 2023-08-09
use channel name or tags to filter `!libs`

cuppachino:
This should mean when you're in tft-dev and run !libs you get a list of tft libraries. Any extra arguments are appended to the topic list instead of replying with an error.

### DIFF
--- a/src/RiotAPILibraries.ts
+++ b/src/RiotAPILibraries.ts
@@ -50,6 +50,13 @@ interface LibraryDescription {
 }
 
 export default class RiotAPILibraries {
+    static CHANNEL_TOPICS = {
+        lol: ["lol", "v4"],
+        lcu: ["lcu"],
+        tft: ["tft"],
+        lor: ["lor"]
+    } as const;
+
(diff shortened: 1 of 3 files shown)


src/RiotAPILibraries.ts
```diff
@@ -50,6 +50,13 @@ interface LibraryDescription {
 }
 
 export default class RiotAPILibraries {
+    static CHANNEL_TOPICS = {
+        lol: ["lol", "v4"],
+        lcu: ["lcu"],
+        tft: ["tft"],
+        lor: ["lor"]
+    } as const;
+
```
Token07, line 59: Maybe consider moving this to shared_settings.json

Token07 (approved):
Seems to work on my test env

src/RiotAPILibraries.ts
cuppachino: good catch 👍

cuppachino:
Is there anything else I can do before merging?

PR #314, opened 2024-01-31
Update to return unmodified response from API

Token07:


Querijn:
I remember now that this was done back in the day because the web frontend used the extra data displayed. This is no longer used, I reckon.

Issue #225, opened 2019-12-08
URL check fails for league entries endpoints


MingweiSamuel:
https://na1.api.riotgames.com/lol/league/v4/entries/RANKED_SOLO_5x5/DIAMOND/I?page=10
https://na1.api.riotgames.com/lol/league-exp/v4/entries/RANKED_SOLO_5x5/DIAMOND/I?page=10

PR #317, opened 2024-03-27
Fix issue #225

Token07:
At last, there's a fix!


Querijn:
I hope you've tested it well enough, because I really don't feel reviewing regex

